PR #9, opened 2022-11-21
Add Server architecture / CQRS section

eliykat:
## Objective

Migrating this PR from the adr repo.

> I found it difficult to follow the CQRS ADR at first, because I haven't used the pattern before, and I wasn't sure how we were implementing it in our codebase. A lot of the information online is fairly high level, which is conceptually helpful, but doesn't always provide much guidance for our specific circumstances.
> 
> This PR aims to add some further guidance for devs who are following this ADR. If I've misunderstood anything, then this is also a great time to correct me ;)
> 
> Upon re-reading the ADR, I also noticed that the title refers to CQRS, but the decision refers to Commands only. I do think that we need queries as well (as separate classes in some cases, not always just the repository call), as there are some operations which are read-only and it would be incongruous to keep using services for them. So I've tweaked this wording to refer to both.


MGibson1:
@eliykat Are you able to convert this over to the architecture section or should we close and/or request help doing that?

eliykat:
@MGibson1 it's on my list for this week :)

eliykat:
Ready for review @MGibson1 .

docs/architecture/server/index.md
```diff
@@ -0,0 +1,119 @@
+---
+sidebar_position: 30
+---
+
+# Server Architecture
+
+## CQRS ([ADR-0008](../adr/0008-server-CQRS-pattern.md))
+
+We are currently transitioning the server to use a Command and Query Responsibility Segregation
```
withinfocus, line 9: ℹ️ I'd like to state things in a more present tense, as this is _what is_ happening per the ADR. ADRs imply change but when referred to as a part of current architecture they "are".

docs/architecture/server/index.md
```diff
@@ -0,0 +1,119 @@
...
+from the database.
+
+**Queries** are read operations, e.g. `GetOrganizationApiKeyQuery`. They should never write to the
+database.
+
+The database is the most common data source we deal with, but others are possible. For example, a
+query could also get data from a remote server.
+
+Each query or command should have a single responsibility. For example: delete a user, get a license
+file, rotate an api key. They are designed around verbs or actions (e.g.
```
withinfocus, line 29: ℹ️ "API".

docs/architecture/server/index.md
```diff
@@ -0,0 +1,119 @@
+---
+sidebar_position: 30
+---
+
+# Server Architecture
```
withinfocus, line 5: ℹ️ Casing of titles is inconsistent. Not something I am hung up for now, but I do want to fix this en masse.

docs/architecture/server/index.md
```diff
@@ -0,0 +1,119 @@
...
+            └── RotateOrganizationApiKeyCommand.cs
+```
+
+### Maintaining the command/query distinction
+
+By separating read and write operations, CQRS encourages us to maintain loose coupling between
+classes. There are two golden rules to follow when using CQRS in our codebase:
+
+- **Commands should never read and queries should never write**
+- **Commands and queries should never call eachother**
```
withinfocus, line 74: ℹ️ "eachother".

docs/architecture/adr/0008-server-CQRS-pattern.md
```diff
@@ -28,9 +28,9 @@ method parameters, which goes against our typical DI pattern.
 ## Considered Options
 
 - **`<<Entity>>Services`** - Discussed above
```
withinfocus, line 30: ℹ️ I recommend double-dashing your separators throughout.

docs/architecture/server/index.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+This has unit testing benefits as well - instead of having lengthy "arrange" phases where you mock
+query results, you can simply supply different argument values using the `Autodata` attribute.
+
+### Avoid Primitive Obsession
+
+Where practical, your commands and queries should take and return whole objects (e.g. `User`) rather
+than individual properties (e.g. `userId`).
+[Read more about Primitive Obsession](https://refactoring.guru/smells/primitive-obsession).
```
withinfocus, line 102: ℹ️ I'm seeing a trend where we name links at the bottom of the file vs. inline, but not sure how consistent this is.

docs/architecture/server/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Avoid excessive optional parameters
+
+Lots of optional parameters can quickly become difficult to work with. Instead, consider using
+method overloading to provide different entry points into your command or query.
+
+### Transition plan
+
+We are gradually transitioning to a CQRS pattern over time. If you're making changes that use or
```
withinfocus, line 111: ℹ️ Similar comment -- best to speak to the intention of what "is" and move and of this transitional content to the ADR.

withinfocus (changes requested):

docs/architecture/server/index.md
eliykat: huh. TIL 😁

docs/architecture/server/index.md
eliykat: Rephrased to remove the reference to the transition

docs/architecture/server/index.md
eliykat: True. I've made all sub-headings sentence case (because they mostly were already) but left this top-level heading using title case (which seems to be consistent across pages). I guess that makes it consistently inconsistent?

docs/architecture/server/index.md
eliykat: I wasn't really sure where to include this link, so it is a bit shoehorned in. I've made the sub-heading an inline link instead.

docs/architecture/adr/0008-server-CQRS-pattern.md
eliykat: Done, this is not generally done elsewhere though, so probably something that also needs a bulk update + linting rule if you feel strongly about it.

docs/architecture/server/index.md
eliykat: My thinking was that most people will not read the ADR (which is basically a historical document), and we need as many devs as possible to contribute to migrating existing logic. So it's useful here for visibility. I've gone ahead and done it anyway though, in case this doesn't convince you.

withinfocus (approved):

PR #24, opened 2022-12-07
Added client service architecture

trmartin4:
## Objective

Moved the Client Service Architecture from [Confluence](https://bitwarden.atlassian.net/wiki/spaces/DEV/pages/240287941/Vision+for+Client+Service+Architecture) to the contributing docs.


docs/architecture/clients/services/implementation.md
```diff
@@ -0,0 +1,200 @@
...
+| `[Domain]Service`         | Provides read access to the domain through an Observable, which is updated when the local state changes via an RxJS `BehaviorSubject`. This service also provides helper methods related to that domain.                                      | `PolicyService`         |
+| `Internal[Domain]Service` | Provides public update methods on the service's Observable properties _without_ updating the `[Domain]` on the server. It is implemented as an abstract class which is extends `[Domain]Service`, but generally not shared with most classes. | `InternalPolicyService` |
+| `[Domain]ApiService`      | Provides synced-with-server write capabilities. In general, if you’re editing a `[Domain]`, you want `[Domain]ApiService`.                                                                                                                    | `PolicyApiService`      |
+
+:::tip Why a separate ApiService for the domain?
```
MGibson1, line 63: This title gets capitalized, which makes it hard to read `ApiService`. I suggest either wrapping as inline code (`ApiService`) or two words (API Service)

docs/architecture/clients/services/implementation.md
```diff
@@ -0,0 +1,200 @@
...
+   client
+
+You can see those responsibilities above, where `ApiService` updates the server and
+`Internal[Domain]Service` updates the `BehaviorSubject`.
+
+:::note
+
+In our current implementation, `Internal[Domain]Service` is implemented by `[Domain]Service`.
+However, that implementation is abstracted from the `[Domain]ApiService` so that there is not a
+direct dependency. This will allow for independent implementations in the future.
```
MGibson1, line 134: To my knowledge, there are no plans to make Interanl impls independent from public ones in the future.

The progress I envision is to make `Internal[Domain]Service` internal to the module with only observables/api calls made public by a modules `index.ts` barrel file

docs/architecture/clients/services/implementation.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+For domain updates from another client, the current client receives those messages through the
+`SyncService`. The `SyncService` takes a dependency on the `Internal[Domain]Service`, as it is
+responsible for updating only the internal cache. The `SyncService` updates the domain through this
+service to propagate the changes in the current client.
+
+The `Internal[Domain]Service` uses `BehaviorSubject`s in the `[Domain]Service` for that caching
+mechanism, but that implementation is abstracted from the `SyncService` and any other consumers of
+`Internal[Domain]Service`.
```
MGibson1, line 173: Not all domains cache, this is an implementation detail of the domain. You point that out, but maybe change `uses` to `may use` to indicate it's not a guarantee.

MGibson1 (changes requested):
I've only reviewed services docs thus far. What do you think about moving push notifications to another PR?

It's kind of different expertise to ensure correctness imo

docs/architecture/clients/services/implementation.md
```diff
@@ -0,0 +1,200 @@
+---
+sidebar_position: 2
+---
+
+# Implementation
+
+Bitwarden's client service architecture is aligned with the [vision](vision.md) that we have
+defined. These goals were the driving force behind the current implementation.
```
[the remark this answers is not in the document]
trmartin4: I attempted to soften this a bit.

docs/architecture/clients/services/implementation.md
```diff
@@ -0,0 +1,200 @@
...
+}
+class ApiService {
+  + send()
+}
+
+"UI Component" --> "[Domain]ApiService"
+"[Domain]ApiService" --> ApiService
+"[Domain]ApiService" --> "Internal[Domain]Service"
+"Internal[Domain]Service" --> "[Domain]Service"
+"[Domain]Service" ..> "Other UI Component" : Observable next()
```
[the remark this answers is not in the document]
trmartin4: I see "Other UI component" on the left now:

image

There is a `together` command that you can use to tell certain nodes to stay near each other on the diagram.  Is that what you were thinking?

docs/architecture/clients/services/implementation.md
trmartin4: Done. Good idea.

docs/architecture/clients/services/implementation.md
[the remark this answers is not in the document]
trmartin4: I agree.  I've removed it.

docs/architecture/clients/services/implementation.md
trmartin4: I removed the callout altogether, since my intuition for the reasoning was completely wrong 😄 .  If you feel it requires clarification I can add something back.

docs/architecture/clients/services/implementation.md
trmartin4: I removed this section.

docs/architecture/clients/services/implementation.md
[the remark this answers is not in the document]
MGibson1: I had meant that UIComponent should be on the left, but I think this is OK, really. It's just a little confusing that the component doing the updating isn't the one that I read first.

MGibson1 (approved):
This looks great. Sorry It's taken so long for me to come back to it. Thank you for doing the translation.

Some things I think we'll want to think through soon
1. We probably want to expand this soon to touch on how DomainServices gain access to state, and make the use of Sync explicitly the method by which the DomainService pattern gains access to data.
2. I think we want to flesh out the direct access to server through `get` methods from UI Components. I haven't done enough thinking about this, but one of my assumed truths was that the presence of an ApiService in a Component meant that it edited data on that Domain. It turns out, isn't true because of these `get` methods. We should think on whether that good, bad, or neutral.

PR #30, opened 2022-12-14
Extended title for ADR 0013

trmartin4:
## Objective

The title for ADR 0013 didn't include "for request/response models". It wasn't clear from the title what the scope of the change was, especially when compared with 0011 ("Angular folder structure").

If there is a hesitancy to extend the ADR title to a third line (shown below), perhaps we could include phrase it as "Request/response model folder structure", as the "request/response model" is important context to have.

image


MGibson1 (commented):
Seems reasonable to me, @Hinton what do you think? IIRC this title is off because you wanted a bigger scope but we limited ourselves

trmartin4:
Perhaps to maintain consistency and clearly delineate the scope, we could title it "Service request/response folder structure" to parallel "Angular folder structure"

Hinton:
Perhaps we could rename it to `Deprecate global request/response models`?

trmartin4:
@Hinton that sounds great!  I've updated it.

Hinton (commented):
We should change the filename while we're at it

trmartin4:
> We should change the filename while we're at it

Done.  I adjusted the title in the document as well.

Hinton (approved):

docs/architecture/adr/0013-deprecate-global-request-response-models.md
```diff
@@ -1,12 +1,12 @@
 ---
-title: 0013 - Avoid layered folder structure
+title: 0013 - Deprecate global request/response models
```
Hinton, line 2: We can remove the title since it's identical to the header below.

Hinton (changes requested):

docs/architecture/adr/0013-deprecate-global-request-response-models.md
Hinton, line 2:
```suggestion
```

Hinton (approved):

PR #37, opened 2022-12-19
Added browser environment setup

trmartin4:
## Objective

Define the different ways that the browser extension environment can be configured during development.


docs/getting-started/clients/browser/index.md
```diff
@@ -8,6 +8,79 @@ sidebar_position: 3
 
 Before you start, you must complete the [Clients repository setup instructions](../index.md).
 
+## Environment Setup
```
differsthecat, line 11: What do you think about putting this section below the `Build Instructions` section? It is pre-configuration, but seems a little more in-depth than the run instructions and can be used after building and running if people wish to do so

docs/getting-started/clients/browser/index.md
```diff
@@ -8,6 +8,79 @@ sidebar_position: 3
...
+
+By default, the browser extension will run pointing to the production server endpoints. To override
+this for local development and testing, there are several options.
+
+### Using `managedEnvironment`
+
+The browser extension has the concept of a "managed environment", which is JSON configuration stored
+in
+[development.json](https://github.com/bitwarden/clients/blob/master/apps/browser/config/development.json).
```
JaredSnider-Bitwarden, line 20: Dev json should probably be wrapped in `` similarly to the BrowserEnvironmentService for consistency.

docs/getting-started/clients/browser/index.md
```diff
@@ -8,6 +8,79 @@ sidebar_position: 3
...
+The `managedEnvironment` setting allows the contributor to override any or all of the URLs for the
```
JaredSnider-Bitwarden, line 21: I feel like we should include the JSON hierarchy somewhere so a reader could know that this setting exists under
the top level `devFlags {...}` JSON object without having to click on the `development.json` link

docs/getting-started/clients/browser/index.md
JaredSnider-Bitwarden: Agreed.

docs/getting-started/clients/browser/index.md
```diff
@@ -8,6 +8,79 @@ sidebar_position: 3
...
+There are two ways to use `managedEnvironment`, depending upon whether you will also be running the
+web vault at the same time.
+
+#### `managedEnvironment` with web vault running
+
+If you are also running the web vault, you only need to set the `base` URL in the
+`managedEnvironment`:
+
+```json
+    "managedEnvironment": {
```
JaredSnider-Bitwarden, line 35: I think including the full JSON structure provides better clarity. 
```json
{
   "devFlags":{
      "managedEnvironment":{
         "base":"https://localhost:8080"
      }
      ...
   }
   ...
}
```

docs/getting-started/clients/browser/index.md
```diff
@@ -8,6 +8,79 @@ sidebar_position: 3
...
+      "icons": "http://localhost:50024"
+    }
+```
+
+### Manually setting the Custom Environment URLs
+
+You may want to adjust the server URLs to point to your local server once you have loaded the
+extension instead of overriding them in `managedEnvironment`. You can change this through the
+browser settings. You can see instructions on how to configure the URLs
+[here](https://bitwarden.com/help/change-client-environment/).
```
JaredSnider-Bitwarden, line 82: Providing an example of the locally configured settings screen might be useful?
image

JaredSnider-Bitwarden (commented):
Overall, great work on this, and thank you for writing this up. This provides much needed additional clarity regarding how the browser environment configuration works! 🎉

docs/getting-started/clients/browser/index.md
trmartin4: Great idea!  Done.

docs/getting-started/clients/browser/index.md
trmartin4: Good idea.  I've included a snipped of the file as well as clarified it in the verbiage.

docs/getting-started/clients/browser/index.md
trmartin4: Good idea.  I've adjusted both cases.

docs/getting-started/clients/browser/index.md
trmartin4: I actually had an image originally, and Oscar pointed out that we had it documented.  But the image you included does also show what it should look like for local development (not just abstractly how to change your Custom Environment for self-hosted).  I've added both, as it does add more context.  Thank you!

JaredSnider-Bitwarden (approved):
Awesome! Thanks again for the write up and changes!

PR #38, opened 2022-12-20
Fix info callout in MAS dev page

differsthecat:
## Objective

I noticed an info callout on the MAS desktop doc is using the old format causing weirdness. Switch to the correct syntax.


JaredSnider-Bitwarden (approved):
Looks great!
Before:
image

After: 
image

PR #44, opened 2023-01-03
Added troubleshooting to sign already committed changes

fedemkr:
## Objective

Added fix for the situation in which you have already committed your changes and you forgot to configure GPG signing

PR #41, opened 2022-12-30
[EC-761] Mobile overview and Apple Watch docs

fedemkr:
## Objective

The main goal of this PR is to add Apple Watch documentation regarding its architecture and how to develop.
Also a tiny overview of the Mobile clients was added.

PR #48, opened 2023-01-10
Added troubleshooting to sign already committed changes

fedemkr:
## Objective

Added fix for the situation in which you have already committed your changes and you forgot to configure GPG signing

_Note: It's a new PR from #44 with all the commits signed given that rebasing there somehow didn't work as I expected (yes it's kind of a paradox given the goal of the PR), maybe for having merges in between or because of targeting a commit in the middle of the PR instead of previous to the beginning of the first PR commit 🤔_ 

_Sorry for calling review again, also added a note to do what I just did with this to fix it without rebasing which is basically create a new PR_

### DIFF
--- a/docs/contributing/commit-signing.mdx
+++ b/docs/contributing/commit-signing.mdx
@@ -101,6 +101,19 @@ Refer to
   your terminal. Refer to
   [this troubleshooting document](https://gist.github.com/paolocarrasco/18ca8fe6e63490ae1be23e84a7039374)
   for more help with this error
+- If you have already committed your changes and forgot to configure gpg signing you can fix it by
+  doing a rebase like the following targeting the hash of the previous commit to the one you did (it
+  will sign all commits from there):
+
+  ```bash
+  git rebase --exec 'git commit --amend --no-edit -n -S' -i <previous-commmit-hash>
+  ```
+
+  Beware that if you had already pushed your changes to origin, you are going to have to do a force
+  push in git after applying the fix which shouldn't be harmful on side branches but one needs to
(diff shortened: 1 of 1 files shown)


Hinton:
I tried running this command on a PR which had merge commits which caused a lot of merge conflicts that had to be resolved. Is this something you've tried @fedemkr?

I've personally primarily squashed my history to match master when I've had to resolve this type of issue.

fedemkr:
Yes, on #41 it worked the last time I did it (where there were just 4 commits, now I see that I need to re run it given that I merged master afterwards before I knew about the SourceTree issue and it has no GPG).
I'll re run it again on that PR and I'll see what I get.
Given that you also had problems running the command maybe this needs further testing to check the conditions in which it works to avoid causing more problems. I'll label it as Hold for now

fedemkr:
Testing on #41, I realized that the target hash should be the **current master commit hash** so when you do the rebase the operation will be performed from there and all the side branch commits will be then signed on top of master which is the expected behavior (so the diff is the one we expect). 
Previously, when I said it worked targeting the previous commit hash to the one in the branch, it was working because it happened to be the hash of the commit of current master.
What I'm not 100% sure yet is if this would work fine with non-automatic merge commits, like you merge from master but you have to perform some manual changes on the merge and commit them. I think it should work fine as well but haven't tested that situation yet.

eliykat:
@fedemkr  This probably isn't needed anymore because we disabled commit signing requirements.

fedemkr:
Thanks @eliykat. Closing this then because it's disabled.

PR #51, opened 2023-01-17
ADR on adoption of Web Push

withinfocus:
## Objective

ADR on how we'll migrate to use of Web Push and a new provider for mobile and server, all vs. SignalR.

### DIFF
--- a/docs/architecture/adr/0016-web-push-sunset-signalr.md
+++ b/docs/architecture/adr/0016-web-push-sunset-signalr.md
@@ -1,0 +1,109 @@
+---
+adr: "0016"
+status: In progress
+date: 2023-01-12
+tags: [server, clients, browser, notifications]
+---
+
+# 0016 - Adoption of Web Push
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+## Context and Problem Statement
+
+Millions of users now utilize push notifications for background vault syncs and passwordless login
+requests. As the platform has grown, so has the need to maintain the current
+[SignalR][signalr]-based solution that utilizes WebSockets, a system that places enormous pressure
+on cloud components in the event of deployments or other application updates. Combined with the need
+to also support mobile devices with proprietary push notification protocols, a next-generation and
+hybridized service offering is needed to simplify and scale up notification delivery.
+
+### Modern Infrastructure Management
+
+The [notifications service][notifications] is an independently-deployed component and for the
+Bitwarden-hosted version is backed by dozens of virtual machines each supporting tens of thousands
+of concurrent connections. With the move to less manual component maintenance in the cloud --
+specifically Kubernetes -- the burden of keeping up the large amount of connections with
+[pods][podlife] is complex and performance spikes are expected.
+
+### WebSockets
+
+Usage of SignalR, even with much larger workloads, has at face value worked well, but the nature of
+the work being performed and usage of persistent WebSockets connections for certain clients has
+become tenuous. In aggregate the amount of input and output from the notifications service is quite
+large and is used almost entirely for background-oriented operations. Various new technologies have
+come out that are better suited towards synchronization work. The technology or protocol used must
+be usable at large scale in the cloud as well as for self-hosted installations.
+
+Browser extension changes and mandates such as [Manifest V3][mv3] additionally present support
+problems with long-lived background connections.
+
+### Cost and Reliability
+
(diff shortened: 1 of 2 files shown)


Hinton:
I think the biggest unknown with web push is how it would work with the web vault, I worry that we'll need another solution than web push there.

- Web push requires user approval to enable.
- We only want notifications while the session is active i.e. as long as the tab is open. Web push is more geared towards a persisted state where you login once and continue to receive notifications in the background while the page might be closed.

docs/architecture/adr/0016-web-push-sunset-signalr.md
```diff
@@ -0,0 +1,109 @@
...
+- **Work around Azure Notification Hubs limits** - Since all mobile devices, even for self-hosted
+  installations, must utilize the Bitwarden cloud for push due to certificate security, devise a
+  solution that shards devices across many Notification Hubs within new subscriptions.
+- **Adopt a new offering for mobile push notifications only** - Use something other than Azure
+  Notification Hubs that doesn't have limits, perhaps with some technology sacrifices.
+- **Adopt a new offering for non-mobile push notifications only** - Use something other than SignalR
+  like a homegrown [Web Push][webpush] backend inside the notifications service.
+- **Adopt a new combined push service provider and utilize Web Push** - Not only migrate away from
+  Azure Notification Hubs but also utilize Web Push with a single, reliable service provider. Host a
+  compatible Web Push backend for self-host with the clients that need it.
```
MGibson1, line 65: Can you expand on hosting a compatible web push backend, does this mean extending our relay notifications to web push or try to find a way for self host to push directly to browser push services?

docs/architecture/adr/0019-adopt-web-push.md
```diff
@@ -0,0 +1,111 @@
...
+- **Work around Azure Notification Hubs limits** - Since all mobile devices, even for self-hosted
+  installations, must utilize the Bitwarden cloud for push due to certificate security, devise a
+  solution that shards devices across many Notification Hubs within new subscriptions.
+- **Adopt a new offering for mobile push notifications only** - Use something other than Azure
+  Notification Hubs that doesn't have limits, perhaps with some technology sacrifices.
+- **Adopt a new offering for non-mobile push notifications only** - Use something other than SignalR
+  like a homegrown [Web Push][webpush] backend inside the notifications service.
+- **Adopt a new combined push service provider and utilize Web Push** - Not only migrate away from
+  Azure Notification Hubs but also utilize Web Push with a single, reliable service provider. Host a
+  compatible Web Push backend for self-host with the clients that need it.
```
Hinton, line 66: I feel like we're essentially deciding two things here. And I wonder if this could be better broken up as two decisions within this ADR.

1. 
    a) We will migrate the browser client to Web Push since it lowers the number of concurrent clients.
    b) Stick with SignalR for browsers

2. 
    a) We will adopt a service provider which offers both mobile and web push notifications.
    b) Find a new vendor offering Web Push
    c) Build our own Web Push system

docs/architecture/adr/0019-adopt-web-push.md
```diff
@@ -0,0 +1,111 @@
...
+- Several service providers offer Web Push backends for our cloud offering, and the protocol can be
+  implemented within the notifications service for self-host.
+- Web Push fits well into Manifest V3 and service workers.
+- Infrastructure maintenance burdens and cost for the notification service should significantly
+  decrease.
+
+### Negative Consequences
+
+- Need to watch Safari support for Web Push which was just recently released at time of writing.
+- Costs for selecting a service provider.
```
Hinton, line 85: Potential cost, but we also have a cost for our current provider. I also believe the provider we found would potentially be less expensive than our current solution.

Hinton (commented):
Looks mostly good, I think we can be a bit more explicit in considered options if we restructure this as two decisions. (The second decision depending on the first)

docs/architecture/adr/0019-adopt-web-push.md
withinfocus: I am indeed throwing a lot into one ADR (deliberately) so they're intertwined to me, but I broke out the topics more clearly.

Hinton (approved):
👍, please try to avoid rebasing in the future. It makes the review process more annoying since GitHub can't keep track of your last position.

PR #52, opened 2023-01-18
ADR on feature management

withinfocus:
## Objective

ADR on how we'll begin to use a feature management service provider.

### DIFF
--- a/docs/architecture/adr/0018-feature-management.md
+++ b/docs/architecture/adr/0018-feature-management.md
@@ -50,7 +50,7 @@ Chosen option: **Adopt a feature management system with local fallback**.
 ### Plan
 
 The [server][server] codebase will adopt a .NET SDK for a service provider that offers feature
-management. Only the serverside SDK will be used at first to manage access and cost and feature
+management. Only the serverside SDK will be used to manage access and cost and feature
 states will be communicated down to calling clients where appropriate via API response elements. New
 features will be set up inside the service provider's platform, and changes to them will be streamed
 to the running applications. Access to the provider will be controlled internally.
--- a/docs/architecture/adr/0018-feature-management.md
+++ b/docs/architecture/adr/0018-feature-management.md
@@ -55,18 +55,29 @@ states will be communicated down to calling clients where appropriate via API re
 features will be set up inside the service provider's platform, and changes to them will be streamed
 to the running applications. Access to the provider will be controlled internally.
 
-Contexts -- which for most use cases represent individual users -- will be established that
-communicate to the API using supported clients. Said contexts will be available within the service
-provider for specific targeting as desired. Integrations with organization and enterprise management
-will be considered at a later date.
+To facilitate when feature states need to be used by clients amongst other configuration, a new API
+will be developed to provide, amongst potentially other things, the server version and other build
+characteristics such as commit hash as well as a collection of configurations values. Some of these
+values are already maintained persistently and will be intermixed with feature keys. Clients will
+refresh configuration upon startup, when their local configuration is updated, and when sync events
+come in.
+
+Contexts -- generally representing individual users -- will be established that communicate to the
+API using supported clients. Said contexts will be available within the service provider for
+specific targeting as desired. Contexts will be established for the user, organization, and
+provider, with unique IDs for the entity as a key and other details as needed. Context attributes
+when needed can be marked as private to avoid spillover to the service provider, and the provider
+will be added to the [subprocessor list][subprocessors] with respective communication.
 
 Compile-time configuration will be converted wherever possible to use the feature management service
-provider.
+provider. SDK access to the service provider will be segmented by environment; some features may
+never progress to all environments.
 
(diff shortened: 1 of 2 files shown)


Hinton:
Some prio work on this topic. https://bitwarden.atlassian.net/wiki/spaces/PIQ/pages/302481409. We also have sort of a venue for communicating flags to the clients, https://bitwarden.atlassian.net/wiki/spaces/PIQ/pages/178782622.

I think there are some unknowns to figure out related to this ADR.
- Example of how feature flags would work on clients.
	- Due to privacy reasons we probably don't want the clients connecting directly to a third party service.
	- Is the config endpoint sufficient, and how often do we need to refresh it?
- How do we handle inconsistencies in enabled features due to sync delays? (between clients, but also between servers assuming we need to cache this somehow)

@joseph-flinn I know we discussed this a few days ago, anything I'm missing?

joseph-flinn:
@Hinton I think the only thing I'd add is the option of using env vars for feature flags on self-host.

withinfocus:
Few notes from the update:
- development of the config endpoint per internal documentation is expected to support the clients
- some of what's proposed in that config endpoint is a bit stale so I'm not sure we need server and environment details, but that can be dealt with during work breakdown
- made clear that the service provider won't connect from clients, plus how it will work in self-host (via the existing property check, to go into an offline mode)

We may send some identifiable characteristics for users and organization to a provider, but that'd be covered under them being a registered subprocessor. Service providers can also treat attributes as private but we'll want to use what we can when targeting is needed; again this can be worked through during breakdown.

I think with the few noted config sync events we should have pretty good coverage, and while the server-side will benefit from a streaming architecture we can use syncs to get feature states quickly.

docs/architecture/adr/0018-feature-management.md
```diff
@@ -0,0 +1,90 @@
...
+to the running applications. Access to the provider will be controlled internally.
+
+To facilitate when feature states need to be used by clients amongst other configuration, a new API
+will be developed to provide, amongst potentially other things, the server version and other build
+characteristics such as commit hash as well as a collection of configurations values. Some of these
+values are already maintained persistently and will be intermixed with feature keys. Clients will
+refresh configuration upon startup, when their local configuration is updated, and when sync events
+come in.
+
+Contexts -- generally representing individual users -- will be established that communicate to the
```
Hinton, line 65: This is a bit unclear, first we mention it being individual users, but a few lines below it's organizations and providers.

docs/architecture/adr/0018-feature-management.md
```diff
@@ -0,0 +1,90 @@
...
+allow for customization. Offline access (also implying there is no connectivity needed outside the
+installation) to feature states via the service provider will be available given the default state
+configuration. Local files can exist to load opt-ins to features for self-hosted installations, and
+said installations will default to an offline mode.
+
+The software development lifecycle will be enhanced to make clear that essentially all feature
+development should be protected with flags.
+
+Support for using an [OpenFeature][openfeature]-compatible interface in the codebase will be
+considered at a later date.
```
Hinton, line 86: I haven't explored how far along open feature is but generally using the open interface from the start tends to result in less vendor lockin.

docs/architecture/adr/0018-feature-management.md
```diff
@@ -0,0 +1,90 @@
...
+- Logging and traceability of who can experience or experiment with a feature.
+
+### Negative Consequences
+
+- Costs for selecting a service provider.
+
+### Plan
+
+The [server][server] codebase will adopt a .NET SDK for a service provider that offers feature
+management. Only the serverside SDK will be used at first to manage access and cost and feature
```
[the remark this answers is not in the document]
withinfocus: "Nothing is forever" which is why I wrote this, but that's fine with me.

docs/architecture/adr/0018-feature-management.md
```diff
@@ -0,0 +1,90 @@
...
+states will be communicated down to calling clients where appropriate via API response elements. New
+features will be set up inside the service provider's platform, and changes to them will be streamed
+to the running applications. Access to the provider will be controlled internally.
+
+To facilitate when feature states need to be used by clients amongst other configuration, a new API
+will be developed to provide, amongst potentially other things, the server version and other build
+characteristics such as commit hash as well as a collection of configurations values. Some of these
```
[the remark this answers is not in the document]
withinfocus: News to me! Some side conversations implied this was new.

docs/architecture/adr/0018-feature-management.md
withinfocus: I removed some confusing wording. While it _is_ about a user as the requestor, multiple contexts will be built. This allows more dynamic targeting to perhaps just one "kind" of context, or even multiple.

Hinton (approved):
I believe all of my concerns have been addressed. Might want to give @MGibson1 a day or two to review it.

MGibson1 (approved):
I agree

PR #57, opened 2023-01-31
Add more troubleshooting to AppleCrypto error

jlf0dev:
## Objective

The AppleCrypto error described in our troubleshooting section of the server seems to be pretty common. There have been other fixes found unrelated to the original fix, so we should include those preliminary steps before moving on to the more involved fix.


differsthecat (approved):
Nice addition! I've run into this before as well and a restart has always fixed it.

PR #60, opened 2023-02-09
Remove commit signing requirements

eliykat:
## Objective

We are no longer requiring commit signing, so we should remove these requirements from our documentation.

I thought about leaving them there for informational purposes, but it's not serving much use and it's just more documentation to maintain.

### DIFF
--- a/docs/getting-started/tools/index.md
+++ b/docs/getting-started/tools/index.md
@@ -39,8 +39,7 @@ environment.
 - [NPM](https://www.npmjs.com/) v8 (included with Node)
 - [Rust](https://www.rust-lang.org/tools/install) - Used for native desktop components
 - [Git](https://git-scm.com)
-  - [Commit signing](../../contributing/commit-signing.mdx) is required for all _Bitwarden
-    contributors_ and is strongly encouraged for _community contributors_.
+  - [Commit signing](../../contributing/commit-signing.mdx) is strongly recommended
 
 ### Mobile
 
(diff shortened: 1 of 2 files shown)


withinfocus (changes requested):
Can we put this behind the Bitwarden developer toggle then? I think it still holds value to keep, at least as a resource for how we internally should set up our profiles / machines.

Signing as a part of a collection of security enhancements we utilize are something I think we should have clearly on display, and I think we'll continue to enhance them.

Hinton:
I think the page should still be visible, and we should advocate that community contributors also sign their commits but not require it.

eliykat:
Fair enough - updated to note that it is recommended rather than required

docs/getting-started/tools/index.md
```diff
@@ -39,8 +39,7 @@ environment.
 - [NPM](https://www.npmjs.com/) v8 (included with Node)
 - [Rust](https://www.rust-lang.org/tools/install) - Used for native desktop components
 - [Git](https://git-scm.com)
-  - [Commit signing](../../contributing/commit-signing.mdx) is required for all _Bitwarden
-    contributors_ and is strongly encouraged for _community contributors_.
+  - [Commit signing](../../contributing/commit-signing.mdx) is strongly recommended
```
withinfocus, line 42: ❓ Spacing issue?

eliykat: I think it's meant to be indented under the "Git" dot point (given that it's about configuring git), is that what you mean?

withinfocus: Yes, so if that's intentional we are good!

withinfocus (approved):

PR #69, opened 2023-03-03
[PM-1321] fix: make it apparent you need dotnet sdk

titanism:
If you do `brew install dotnet` you will run into issues with `dotnet restore` command.

To resolve this, you have to install version 6.x from <https://github.com/isen-ng/homebrew-dotnet-sdk-versions#dotnet-sdk-versions-tap>.

We recommend to add this into the docs.  For now, we've simply bolded it.

### DIFF
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -17,7 +17,7 @@ Before you start: make sure you’ve installed the recommended
 - Docker Desktop
 - Visual Studio 2022
 - Powershell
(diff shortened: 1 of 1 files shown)


MGibson1 (commented):
I don't see why this requirement is any more required than the others. Why bold just this one?

Hinton:
Thanks for the suggestion. I think this is clear enough as is though and it's odd to just emphasis a single prerequisite.

PR #74, opened 2023-03-22
Add information about SSL certificates

Hinton:
## Objective

We should guide contributors towards setting up their own certificates to avoid browser warnings, and ensure WebAuthn works.

### DIFF
--- a/docs/getting-started/clients/web-vault/index.mdx
+++ b/docs/getting-started/clients/web-vault/index.mdx
@@ -22,7 +22,7 @@ support WebAuthn which is blocked on websites with certificate errors. Please fo
 [mkcert installation instructions](https://github.com/FiloSottile/mkcert#installation).
 
 ```bash
-# Install CA
+# Generate and install root CA
 mkcert --install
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/clients/web-vault/index.mdx
```diff
@@ -11,10 +11,24 @@ import CodeBlock from "@theme/CodeBlock";
...
 
 - Before you start, you must complete the [Clients repository setup instructions](../index.md).
-
 - If you want to develop against a local server instance, see the
   [Server Setup Guide](../../server/guide.md).
 
+### SSL Certificate
+
+It's recommended to generate a valid SSL certificate for local development rather than using the
+shared development cert. This will prevent browser warnings about invalid certificates and properly
```
MGibson1, line 20: The shared cert is bitwarden specific, right? we should make this section agnostic or have two versions where only bitwarden one references this.

Hinton: No it's checked in to the repository and is used by anyone.

MGibson1 (approved):

PR #86, opened 2023-04-03
[PM-1739] Added GIF demonstrating the creation of a connection between SQL container and Azure Data Studio

jacobHancock0402:
I thought the section instructing the reader on how to connect the SQL container was a little underdeveloped, so I added a GIF to make it more clear.

Screenshot of the GIF


Hinton:
Hi, thanks for the contribution. I don't believe we need this gif and we'll most likely rewrite that section to be less opinionated on db tool.

PR #88, opened 2023-04-10
Accessibly docs tweak

danielleflinn:
## Updated guidelines around using `aria-haspopup`

`aria-haspopup` is not fully supported for dialogs yet. See https://a11ysupport.io/tech/aria/aria-haspopup_attribute


differsthecat (commented):
@danielleflinn It looks like there's a linting error probably caused by the updated prettier rules we did a week or two ago.

I can fix it if you'd like, or if you'd like to take a stab at it the process should be:
1. Run `npm ci` in the terminal from the project root
2. Run `npm run lint` in the terminal from the project root 
3. Fix issues identified in this file and commit them.

differsthecat (approved):
Looks good, thank you!

PR #90, opened 2023-04-17
Restructure architecture navigation

Hinton:
## Objective

Reorganizes the architecture section to be more logically grouped. Puts Architecture back to the top, with ADRs following before client & server architecture, and rounding out with a new category for deep dives.

Before:

image

After:

image


MGibson1 (approved):
I like it, thanks!

PR #97, opened 2023-04-22
Update feature flag documentation

trmartin4:
## Objective

Updated feature flag documentation to include client implementations for web and mobile clients.

Also added a section for self-hosted considerations.


docs/contributing/featureflags.md
```diff
@@ -21,15 +21,63 @@ highlights:
...
+`FeatureFlags` enum.
+
+Once that is defined, the value can be retrieved by injecting the `ConfigService` and using one of
+the retrieval methods:
+
+- `getFeatureFlagBool()`
+- `getFeatureFlagString()`
+- `getFeatureFlagNumber()`
+
+Each of these methods accept a default value. This value should always default to "off".
```
withinfocus, line 58: 💭 This is repeated a few times and I am wondering what the right level of that is.

docs/contributing/featureflags.md
```diff
@@ -21,15 +21,63 @@ highlights:
...
+
+#### Mobile
+
+The feature flag values are retrieved through the `GetAsync()` method on the `ConfigService`. They
+are refreshed from the server at the following points in the application lifecycle:
+
+- The first time a configuration value is accessed
+- Every hour after first access
```
withinfocus, line 66: ❓ No mobile sync integration via push?

docs/contributing/featureflags.md
```diff
@@ -21,15 +21,63 @@ highlights:
...
+
+The feature flag values are retrieved through the `fetchServerConfig()` method on the
+`ConfigService`. They are refreshed from the server at the following points in the application
+lifecycle:
+
+- On application startup
+- Every hour after application startup
+- On sync (both automatic and manual)
+
+To use a feature flag, you should first define the new feature flag as an enum value in the
```
withinfocus, line 48: ❓ I linked to the server constants below, once. Should that be removed or we try to put some links in these spots too?

withinfocus (commented):
Relatively minor stuff -- great improvements to this!

docs/contributing/featureflags.md
trmartin4: I agree.  I think the callout above is enough, so I removed this line.

docs/contributing/featureflags.md
```diff
@@ -127,3 +155,15 @@ When defining the subtasks of a story be sure to include a technical debt cleanu
 of the feature flag from code – it’s essential that these not be left around for too long and assume
 a permanent existence. Address the technical debt at a later phase once the feature launches
 successfully.
+
+### Self-hosted considerations
```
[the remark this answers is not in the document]
trmartin4: I added a callout section above about self-hosted and the local flag file.  How do you like that approach?

docs/contributing/featureflags.md
trmartin4: Added links.

docs/contributing/featureflags.md
trmartin4: André had missed adding it on sync on mobile, so that's been addressed.

docs/contributing/featureflags.md
```diff
@@ -19,35 +19,44 @@ highlights:
...
+| QA Cloud             | LaunchDarkly QA                                        |
+| Production Cloud     | LaunchDarkly Production                                |
 
-1. Inject `IFeatureService` where you need a feature flag. Note that you’ll also need
-   `ICurrentContext` when accessing the feature state.
-2. Check the
-   [FeatureFlagKeys](https://github.com/bitwarden/server/blob/master/src/Core/Constants.cs) constant
-   list for the key you plan on using, and add it if not found.
-3. Utilize the above key constant with the appropriate method on the feature service:
+:::note Self-hosted support
```
withinfocus, line 37: 💭 on what level this should be, note or warning. I think this is fine though.

docs/contributing/featureflags.md
```diff
@@ -57,63 +66,127 @@ configuration too via `FlagDataFilePath`) with:
...
 
-Talk with your team, manager / lead, and product about how and when to create new feature flags.
-Depending on how a project / solution is being delivered it may be best to use the online
-LaunchDarkly experience for setup. That said, LaunchDarkly is integrated into our Jira instance to
-make flag creation
-[easy](https://docs.launchdarkly.com/integrations/jira#creating-a-new-feature-flag-from-a-jira-issue).
+### Definition in LaunchDarkly
+
+In order to test the feature flag in any deployed environment, it must first be defined in the
+LaunchDarkly web app. To do this, request the flag from your Engineering Manager - they will have
```
withinfocus, line 112: ⛏️ I prefer double-dashes for separation.

docs/contributing/featureflags.md
```diff
@@ -57,63 +66,127 @@ configuration too via `FlagDataFilePath`) with:
...
+the appropriate access. You should discuss:
+
+- The data type of the flag
+- The default value of the flag
+- The possible values of the flag (for non-boolean types)
+- Any context-based rules that should drive flag behavior
```
withinfocus, line 118: ⛏️ I like to end bullets in parentheses (and there's another example below) but this is rather inconsistent across the docs.

docs/contributing/featureflags.md
```diff
@@ -57,63 +66,127 @@ configuration too via `FlagDataFilePath`) with:
...
-Consider using flags frequently with feature development. Per the above usage of a local file,
-feature flags don’t have to exist in LaunchDarkly for initial development – **don’t create them
-online until you’re sure about the final implementation**. While there is no precise rule on what
-constitutes a "feature", work together on the best balance of flags and their respective purposes.
+### Server
+
+1. Inject `IFeatureService` where you need a feature flag. Note that you’ll also need
+   `ICurrentContext` when accessing the feature state.
+2. Check the
+   [FeatureFlagKeys](https://github.com/bitwarden/server/blob/master/src/Core/Constants.cs) constant
```
withinfocus, line 181: ⛏️ Your other links treat this stylistically as `code` so we should convert this too.

docs/contributing/featureflags.md
```diff
@@ -19,35 +19,44 @@ highlights:
...
-maintained should a flag be unavailable altogether. When an interface supports it, also provide
-default values implying "off" to feature flag accessors.
+When consuming feature flags in either the client or server code, it is important to understand
+where the flags are sourced.
 
-Offline mode makes default values even more important, and local development as well as self-hosted
-installations imply being offline. Set a safe default value not just in the flag definition online
-at LaunchDarkly but in code.
+The source of the flags is dependent upon the Bitwarden server instance that is being used, as for
+client development the flags are served from the Bitwarden `Api`.
```
withinfocus, line 28: ⛏️ This API reference doesn't seem like it needs to look like `code`.

withinfocus (changes requested):
A few really minor things that I hope only take a moment to resolve.

docs/contributing/featureflags.md
trmartin4: I was back and forth on that as well, but changing it to Warning does seem appropriate.  I've changed it.

docs/contributing/featureflags.md
trmartin4: I think you meant period, right?  If so, I've added them.

trmartin4:
@withinfocus I also updated the documentation around where the `flags.json` file needs to be placed.  Gbubemi experimented with it, and you can place it in the project directory instead of the build directory and it still works (perhaps because it is picked up as build content?).  It seems easier to direct people to put it there, but I'd welcome your opinion on that.

docs/contributing/featureflags.md
```diff
@@ -66,10 +69,8 @@ following:
 }
 ```
 
-2. Replace `someKey` with your key name and set the value.
-3. Place the file in the build output directory of the server project that will be consuming the
-   flag. Ensure the file is there before starting the build, but you can change the file contents
-   and see immediate results in running / debugging code.
+Ensure the file is there before starting the build. Once there, you can change the file contents and
```
withinfocus, line 72: ❓ Should this be 3? Or we remove 1 and 2 then?

trmartin4:
> Found a couple minor things. Before we change this though, can you test with non-web applications and the file usage? I think by its nature projects like the API can pull in found files and I guess LD is figuring that out, but I don't anticipate that happening for something like a console app.

I tested it with `EventsProcessor` and saw the same behavior - `flags.json` in the project base directory ended up with the file in `src/EventsProcessor/bin/debug/net6.0/`

withinfocus (approved):
Love it! Thanks for all the great updates.

I'll be looking into alternative data stores in the background.

PR #104, opened 2023-04-26
Update Safari Preferences to Settings

jprusik:
## Objective

It appears Safari currently uses "Settings" terminology to refer to the documented menu options.

Screenshot 2023-04-26 at 3 01 21 PM


gbubemismith (approved):
Thanks for updating this @jprusik

jprusik:
I just noticed this appears to be the case for Firefox and Chrome too (terminology enforced OS-wide?)

gbubemismith:
> Thanks for updating this @jprusik



> I just noticed this appears to be the case for Firefox and Chrome too (terminology enforced OS-wide?)

We might want to check how it appears on non-MAC OS for the other browsers

jprusik:
> 

On Windows 11 both Firefox and Chrome use "Settings" 👍

gbubemismith (approved):
This looks good!

PR #115, opened 2023-05-01
Doc Clarifications

jprusik:
## Objective

Clarify documentation for:

- initial SSO setup
- determining the correct Xcode version to use for Xamarin.iOS

### DIFF
--- a/docs/getting-started/clients/mobile/ios/index.mdx
+++ b/docs/getting-started/clients/mobile/ios/index.mdx
@@ -88,8 +88,8 @@ your version of Xamarin may not yet support the version of Xcode that you are us
 ![](./troubleshoot-generic-simulator.png)
 
 To work around this issue, try [downloading](https://developer.apple.com/download/all/) and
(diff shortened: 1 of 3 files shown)


docs/getting-started/server/sso/index.md
```diff
@@ -11,6 +11,14 @@ You can set up SSO for development using:
 
 - [a local IdP](./local.md) (recommended)
 - [Okta](./okta.mdx)
+- a local server running the SSO project, e.g.
```
gbubemismith, line 14: It looks like this is handled [here](https://github.com/bitwarden/contributing-docs/blame/5b841fc8c7fdc98de6a19a80aa6677d3a3354613/docs/getting-started/server/sso/local.md) so this block can be removed.

gbubemismith (commented):
This is looking good; just an observation

gbubemismith (approved):
Thanks for making the changes. This looks better.

PR #117, opened 2023-05-04
Clarified feature flagging server constant definition

trmartin4:
## Objective

Added section to clarify that a feature flag is required to be defined on the server before use.

### DIFF
--- a/docs/contributing/featureflags.md
+++ b/docs/contributing/featureflags.md
@@ -27,14 +27,14 @@ where the flags are sourced.
 The source of the flags is dependent upon the Bitwarden server instance that is being used, as for
 client development the flags are served from the Bitwarden API.
 
-| Server configuration | Flag source                                    |
-| -------------------- | ---------------------------------------------- |
-| Local development    | Local JSON file                                |
-| Self-hosted          | Flags are "off", unless local JSON is provided |
-| QA Cloud             | LaunchDarkly QA                                |
-| Production Cloud     | LaunchDarkly Production                        |
+| Server configuration | Flag source                                         |
+| -------------------- | --------------------------------------------------- |
+| Local development    | Local JSON file                                     |
+| Self-hosted          | Flags are "off", unless local JSON file is provided |
+| QA Cloud             | LaunchDarkly QA                                     |
+| Production Cloud     | LaunchDarkly Production                             |
 
-:::warning Self-hosted support
+:::caution Self-hosted support
 
 Feature flags are not officially supported for self-hosted customers. Using a local JSON file is not
 a supported method of sourcing feature flag values, outside of Bitwarden internal testing. See
--- a/docs/contributing/featureflags.md
+++ b/docs/contributing/featureflags.md
@@ -91,6 +91,7 @@ Recommendations for naming are:
 
 - Name the flag using kebab-case (lowercase and dash-separated, such as `enable-feature`).
 - Name the flag in the affirmative when possible (`enable-feature`, not `disable-feature`).
+- Keep key names succinct.
 
 Once a name has been decided, add the feature flag to the
 [`FeatureFlagKeys`](https://github.com/bitwarden/server/blob/master/src/Core/Constants.cs) constants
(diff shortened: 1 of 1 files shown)


docs/contributing/featureflags.md
```diff
@@ -57,7 +57,8 @@ see immediate results in running / debugging code.
 To set up the local file data store:
 
 1. Create a file named `flags.json` file in the root project directory (e.g. `/src/Api/` for the
-   `Api` project). The name and path can be changed in configuration via `FlagDataFilePath`.
+   `Api` project). The name and path can be changed via the
+   `globalSettings__LaunchDarkly__FlagDataFilePath` environment variable.
```
withinfocus, line 61: ❓ How often is this variable formatting revealed in contributing docs? It implies a runtime environment vs. what's in the application settings themselves and the LHS here is "enough" to me to explain what to change.

withinfocus (changes requested):
Some expansion on naming style is the one true requested change.

docs/contributing/featureflags.md
```diff
@@ -86,16 +86,24 @@ feature flag. The team should agree on the scope of what is flagged and where th
 applied - both client-side and server-side. While there is no precise rule on what constitutes a
 "feature", work together on the best balance of flags and their respective purposes.
 
+Once you have decided that a feature flag is necessary, the first step is to decide on a name.
+Recommendations for naming are:
```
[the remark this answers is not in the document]
trmartin4: My concern is that we're asking someone to use `flags.json` as a starting point, and then asking them to put the flag name in the server code.  Throughout that process they wouldn't have any LD interaction, or any indication of how LD prefers to structure the key name.  Then, when moving into LD proper, the developer would have to change the server code (and client as well, potentially) with the new name that LD suggests.  For example, going from `show_feature` or `showFeature` to `show-feature`.  I was trying to avoid that by adding best practices at this step.

I do think it makes sense to pull back in the naming guidance that you had as well, but that is my thinking for being a bit more prescriptive.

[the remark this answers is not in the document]
trmartin4: It also looks like a lot of those bullets referenced the idea that we were going to set up the flags directly in Jira, whereas now the instructions are for someone to reach out to their EM for setup directly in LD.  Do we still need to explicitly tie it to a Jira issue somehow?

docs/contributing/featureflags.md
```diff
@@ -86,16 +86,24 @@ feature flag. The team should agree on the scope of what is flagged and where th
...
+
+- Name the flag using kebab-case (lowercase and dash-separated, such as `enable-feature`).
+- Name the flag in the affirmative when possible (`enable-feature`, not `disable-feature`).
+
+Once a name has been decided, add the feature flag to the
+[`FeatureFlagKeys`](https://github.com/bitwarden/server/blob/master/src/Core/Constants.cs) constants
+file on the server. This will allow the flag to be retrieved from LaunchDarkly via whichever data
+source you configure below.
```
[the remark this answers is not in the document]
trmartin4: Yes, I agree that verbiage should change.  I've updated it to reference back to this section.

docs/contributing/featureflags.md
withinfocus: I think I am good otherwise, but what about this one?

trmartin4: Sorry, I overlooked this one.  See if it looks better now.  I also tried to pare down where I was explicitly referring to the file being named `flags.json`, as that is just the default file name.  I referred to it more generically as the `local data store` or `local JSON file`.

docs/contributing/featureflags.md
```diff
@@ -27,50 +27,51 @@ where the flags are sourced.
...
-| Server configuration | Flag source                                            |
-| -------------------- | ------------------------------------------------------ |
-| Local development    | Local `flags.json` file                                |
-| Self-hosted          | Flags are "off", unless local `flags.json` is provided |
-| QA Cloud             | LaunchDarkly QA                                        |
-| Production Cloud     | LaunchDarkly Production                                |
+| Server configuration | Flag source                                    |
+| -------------------- | ---------------------------------------------- |
+| Local development    | Local JSON file                                |
+| Self-hosted          | Flags are "off", unless local JSON is provided |
```
withinfocus, line 33: ⛏️ JSON ... _file_.

docs/contributing/featureflags.md
```diff
@@ -27,50 +27,51 @@ where the flags are sourced.
...
+| QA Cloud             | LaunchDarkly QA                                |
+| Production Cloud     | LaunchDarkly Production                        |
 
 :::warning Self-hosted support
```
withinfocus, line 37: 💭 I never know which of these is best. I saw a "caution" on another PR and wonder if that is a better fit now.

withinfocus (dismissed):
Looks great and I think reads more clearly now.

docs/contributing/featureflags.md
trmartin4: I like that!  I've updated it.  I didn't realize caution was an option.

withinfocus (approved):

PR #126, opened 2023-05-26
Feature flag configuration via application settings

withinfocus:
## Objective

Describes usage of application settings for feature flag configuration via what was added in https://github.com/bitwarden/server/pull/2963.


trmartin4 (approved):
Looks good!  Thank you for updating this and for the `appSettings` changes.

PR #127, opened 2023-05-28
Updated broken link to desktop client docs

trmartin4:
## Objective

Updated link to desktop client docs that still referenced our old documentation URL.


djsmith85 (approved):
Good catch

PR #142, opened 2023-06-27
Update pull request docs

MGibson1:
## Objective

General sprucing-up of the PR contributing section. Includes
* improved criteria for long lived branches
* updated reviewer guidelines due to code owners files
* improved code review tips and requirements

### DIFF
--- a/docs/contributing/pull-requests/branching.md
+++ b/docs/contributing/pull-requests/branching.md
@@ -63,9 +63,7 @@ Choose a **Short-Lived Feature Branch** if the feature will:
 :::tip Still Unsure?
 
 If in doubt, lean toward creating shorter-lived feature branches directly off of `master` for each
-developer’s work, as there is overhead built in to the Long-Lived Feature Branch that can be avoided
-if smaller features can be tested and released independently, or hidden behind a feature flag until
-the changes are ready.
+developer’s work, as there is overhead built in to the Long-Lived Feature Branch.
 
 :::
 
--- a/docs/contributing/pull-requests/branching.md
+++ b/docs/contributing/pull-requests/branching.md
@@ -19,8 +19,9 @@ parts, separated by a slash:
 
 - The team name or abbreviation (e.g. `Vault`), and
 - The Jira issue tag (e.g. `PM-1234`)
+- A short description of the work being done (e.g. `update-csp-hashes`)
 
-In this example, the full branch name would be `Vault/PM-1234`.
+In this example, the full branch name would be `Vault/PM-1234/update-csp-hashes`.
 
 ### Branches for multiple Jira issues
 
@@ -51,22 +52,20 @@ Product.
 
 Choose a **Long-Lived Feature Branch** if the feature will:
 
-- Have multiple authors
-- Consist of multiple Pull Requests, which have already been reviewed
-- Require these multiple Pull Requests from multiple authors in order to produce the testable,
-  releasable change
+- Require multiple Pull Requests to produce a testable, releasable change, and
+- It is not possible to put the changes behind a feature flag
 
 Choose a **Short-Lived Feature Branch** if the feature will:
 
-- Have a single author
-- Consist of a single Pull Request from the Feature Branch into `master` to produce the testable,
-  releasable change
+- Require a single Pull Request to produce a testable, releasable change, or
+- It is possible to put the changes behind a feature flag while multiple PRs are in flight.
 
 :::tip Still Unsure?
 
 If in doubt, lean toward creating shorter-lived feature branches directly off of `master` for each
 developer’s work, as there is overhead built in to the Long-Lived Feature Branch that can be avoided
-if smaller features can be tested and released independently.
+if smaller features can be tested and released independently, or hidden behind a feature flag until
+the changes are ready.
 
 :::
 
@@ -91,8 +90,10 @@ the Long-Lived Feature Branch for the piece of functionality and create a **draf
 branch into `master`. This name should include the Jira Epic name if applicable.
 
 Each developer should then branch off of that feature branch, creating an "Issue Branch" named with
-the initiating Jira issue (e.g. `PM-1234`). We call this an Issue Branch because Jira refers to
-stories and tasks as issues and to differentiate from the Long-Lived Feature Branch above.
+the initiating Jira issue, as described in
+[branches for a specific Jira issue](#branches-for-a-specific-jira-issue). We call this an Issue
+Branch because Jira refers to stories and tasks as issues and to differentiate from the Long-Lived
+Feature Branch above.
 
 :::note
 
--- a/docs/contributing/pull-requests/code-review.md
+++ b/docs/contributing/pull-requests/code-review.md
@@ -37,8 +37,8 @@ from someone else, and there’s nothing wrong in having two reviewers on a PR.
 
 <a id="assumptions-note"></a> When reviewing code, remember that all software is built to conform to
 a set of assumptions. Features, bug fixes, and other requirement changes represent a change in those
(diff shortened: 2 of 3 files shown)


docs/contributing/pull-requests/branching.md
```diff
@@ -51,22 +52,20 @@ Product.
...
+- Require a single Pull Request to produce a testable, releasable change, or
+- It is possible to put the changes behind a feature flag while multiple PRs are in flight.
 
 :::tip Still Unsure?
 
 If in doubt, lean toward creating shorter-lived feature branches directly off of `master` for each
 developer’s work, as there is overhead built in to the Long-Lived Feature Branch that can be avoided
-if smaller features can be tested and released independently.
+if smaller features can be tested and released independently, or hidden behind a feature flag until
+the changes are ready.
```
Hinton, line 68: Feels like we're duplicating the same words as on line 61.

docs/contributing/pull-requests/code-review.md
```diff
@@ -35,6 +33,15 @@ always welcome! For example, it’s okay to leave some general comments or feedb
 that you don’t have enough knowledge to approve the changes. The author can ask for another review
 from someone else, and there’s nothing wrong in having two reviewers on a PR.
 
+:::info Note
+
+<a id="assumptions-note"></a> When reviewing code, remember that all software is built to conform to
+a set of assumptions. Features, bug fixes, and other requirement changes represent a change in those
+assumptions. Code, after merge, should represent the best solution that fulfills requirements, which
```
Hinton, line 40:
```suggestion
assumptions. Code, after merge, should represent the best solution that fulfills the new set of requirements, which
```

docs/contributing/pull-requests/branching.md
MGibson1: Agree, truncated

Hinton (dismissed):

docs/contributing/pull-requests/branching.md
```diff
@@ -19,8 +19,9 @@ parts, separated by a slash:
 
 - The team name or abbreviation (e.g. `Vault`), and
 - The Jira issue tag (e.g. `PM-1234`)
+- A short description of the work being done (e.g. `update-csp-hashes`)
 
-In this example, the full branch name would be `Vault/PM-1234`.
+In this example, the full branch name would be `Vault/PM-1234/update-csp-hashes`.
```
withinfocus, line 24: 💭 What's the benefit of adding this vs. just using a task ID? Our PR template now expects a link to the task and it'll be linked inside Jira. Might be better to _remove_ the task ID here?

docs/contributing/pull-requests/branching.md
```diff
@@ -51,22 +52,18 @@ Product.
 
 Choose a **Long-Lived Feature Branch** if the feature will:
 
-- Have multiple authors
-- Consist of multiple Pull Requests, which have already been reviewed
-- Require these multiple Pull Requests from multiple authors in order to produce the testable,
-  releasable change
+- Require multiple Pull Requests to produce a testable, releasable change, and
+- It is not possible to put the changes behind a feature flag
```
withinfocus, line 56: 🎨 I would enjoy even more language about how long-lived branches are to be avoided and worth a discussion if someone thinks one is needed.

docs/contributing/pull-requests/index.md
```diff
@@ -88,43 +88,23 @@ contributor to revise their local history before pushing to the remote repositor
...
 
-</bitwarden>
-
-### Force Pushing
+## Creating a Pull Request
 
-Once a PR has been reviewed, please avoid git operations that affects the existing git commits, i.e.
-anything that requires a force push. Force pushing prevents GitHub from correctly identifying “new
-changes” to a PR forcing the reviewer to start over again.
+The Bitwarden repositories have a _Pull Request template_ which should be followed. This will ensure
```
withinfocus, line 102: 🎨 Could link to https://github.com/bitwarden/template/blob/main/.github/PULL_REQUEST_TEMPLATE.md.

docs/contributing/pull-requests/branching.md
eliykat: I wasn't going to, but I'll bite: why not just use the auto-generated jira branch name? Then we just don't have to worry about it. Not sure if I've missed the boat on this discussion, but this seems tedious to follow and enforce.

docs/contributing/pull-requests/index.md
MGibson1: good call, I also want to update the template to link here.

docs/contributing/pull-requests/branching.md
MGibson1: @withinfocus the task id in the branch is the means by we can guarantee jira detects a branch belongs to a given ticket. There are other ways, but they don't generate easily searchable branch trees. I also want to be able to understand the gist of what a branch is doing by looking at the name, not go into another tool/web page to determine that. Finally, the description is last to allow auto-completing on the unique ticket.

@eliykat Jira creates branches something along the lines of "board_ticket_description", but the visualization range given to tickets is FAR more than is typical for a branch. Underscores are also not interpreted as separators for branch groups, and it produces the same branch name for each ticket, but we want to allow multiple per ticket if that's appropriate. In short: I just think Jira does a bad job at this.

docs/contributing/pull-requests/branching.md
MGibson1: Open to suggestions or further refinement. 

The way I'm thinking about it is that this sums up the only reason to _have_ a long lived branch is because you need to keep it out of the main branch to be safe. The tip below makes it clear that you should avoid them.

docs/contributing/pull-requests/index.md
MGibson1: actually, we haven't yet standardized on that template and this needs to apply across our repos. I think that addition is more appropriate once that template is emblematic of what you see when opening a PR to one of our production repositories

Hinton (approved):

docs/contributing/pull-requests/branching.md
Hinton: Using the team identifier first allows you to filter out 90% of the branches if your tool supports it. 
image

withinfocus: What's the reason for browsing branches? I guess I don't understand the basis.

docs/contributing/pull-requests/branching.md
withinfocus: Fair enough, and this may be moot in practice.

docs/contributing/pull-requests/index.md
withinfocus: This is currently set up to be the template repository for new repositories in our organization. A collection of engineers reviewed the new PR template and approved it a couple months ago. If there's a desire for further changes it can be done over in the repository, but I do consider it the official template at this point. It's always open for improvements.

docs/contributing/pull-requests/branching.md
MGibson1: I regularly have a bunch of PR reviews in flight, I use branch browsing to keep by local repos organized.

docs/contributing/pull-requests/index.md
MGibson1: I'm not saying it's not official, just that when you open a PR, you don't see that template today. We can (and should) update this to link when we have updated those templates

withinfocus: Well, do you mean when it's everywhere? It's in a good number of repos so far. I'm fine with not linking to the template repo regardless, as you'll use a PR template in all cases.

PR #123, opened 2023-05-19
Added RBAC documentation

trmartin4:
## Objective

Added documentation for Bitwarden Portal RBAC.  I also changed the name from Admin Portal to Bitwarden Portal to reflect current terminology.

PR #144, opened 2023-06-29
Add callout regarding Stripe test subscriptions

shane-melton:
## Objective

Add a callout regarding Stripe test subscriptions and how they [are automatically removed by Stripe](https://support.stripe.com/questions/test-mode-subscription-data-retention) after a certain amount of time. Includes steps to re-create a new Stripe subscription for an existing organization/user using the Bitwarden Portal.


docs/getting-started/server/advanced-setup.md
```diff
@@ -26,6 +26,23 @@ without actually being charged.
...
+
+Stripe has a [policy](https://support.stripe.com/questions/test-mode-subscription-data-retention) to
+**automatically cancel** test subscriptions after 90 days and then **delete** cancelled test
+subscriptions after 30 days. This can cause unexpected Billing behaviors for long-lived premium
+users/organizations in your local server after the subscription is removed.
+
+To correct this, you must re-subscribe the organization/user to a premium plan to create a new test
+subscription.
+
+1. From the Bitwarden Portal, remove the organization/user gateway information and set their plan to
```
eliykat, line 39: I believe Admin Portal is the proper name:

```suggestion
1. From the Admin Portal, remove the organization/user gateway information and set their plan to
```

docs/getting-started/server/advanced-setup.md
```diff
@@ -26,6 +26,23 @@ without actually being charged.
...
+   "Free".
+2. From the web client, add a new test payment method to the organization/user.
+3. Re-purchase the desired premium features as you would normally.
```
eliykat, line 42: Style: you don't need full stops at the end of numbered items

```suggestion
   "Free"
2. From the web client, add a new test payment method to the organization/user
3. Re-purchase the desired premium features as you would normally
```

eliykat (changes requested):
Thanks for figuring this out! I always just created a new org 😁 

These suggestions might need linting to be re-run to fix line lengths.

docs/getting-started/server/advanced-setup.md
shane-melton: This could be outdated, but I did see there is an [open PR](https://github.com/bitwarden/contributing-docs/pull/123#issue-1717746050) renaming this to "Bitwarden Portal" in the docs. 

@trmartin4 Do you know which name is preferred going forward?

trmartin4: I checked with Product, and it is a bit more nuanced than I thought.  The cloud version (which we use internally to manage our cloud-hosted instances) is called "Bitwarden Portal", but the same code running on a self-hosted instance is called the "System Management Portal".  It's the same code (albeit with some `isSelfHosted()` checks that disable some functionality like RBAC), but it has different names from a Product Portfolio perspective.

I got this clarification from Product.  Given that, I _think_ the portal referenced in this change would be the Bitwarden Portal.

To clarify this, I'm going to open another PR and add a callout on the contributing docs page to make sure that this is understood.

eliykat: Thanks for clarifying!

eliykat (approved):

PR #151, opened 2023-07-07
ADR on logging and observability

withinfocus:
## Objective

ADR on refactoring of logging and embracing OpenTelemetry.


Hinton (commented):
Thoughts on splitting this into two ADRs? It feels like two portions of work, that we want to track individually.

1. Standardize logging to stdout & File through serilog.
2. Adopt OpenTelemetry for observability.

withinfocus:
After office hours, closing in favor of two ADRs for logging and observability.

PR #154, opened 2023-07-12
add devcontainer setup docs

tangowithfoxtrot:
## Objective

This adds documentation for VS Code Dev Containers, added by the [associated PR in server](https://github.com/bitwarden/server/pull/3080).


tangowithfoxtrot:
I am unsure why the build workflow is having issues with the link used in the docs; the link works locally and is a direct copy from the standard "Setup Guide" steps.

docs/getting-started/server/guide.md
```diff
@@ -22,6 +22,13 @@ Before you start: make sure you’ve installed the recommended
 
 :::
 
+:::info
+
+A guide for developing in VS Code Dev containers can be found [here](./guide-vscode) if you prefer
```
withinfocus, line 27: ⛏️ Capital "C" for "Containers".

This is more of my style I suppose, but I try to future-proof this a bit with something like a list format. This could instead be reworded as a collection of guides available, and then you might want to have this in a subfolder of "guides". We have VS Code and VS material here and perhaps they should follow your pattern here so it isn't unique and on its own.

I realize that's easy to ask, but it'd be a nice structure change.

docs/getting-started/server/guide-vscode.md
```diff
@@ -0,0 +1,267 @@
...
+After you’ve run the deployed the Dev Containers, you can use the
+[Docker Dashboard](https://docs.docker.com/desktop/dashboard/) or `docker` CLI to manage your
+containers. You should see your containers running under the `bitwarden_common` group.
+
+:::caution
+
+Changing `MSSQL_SA_PASSWORD` variable after first running the Dev Containers will require a
+re-creation of the storage volume.
+
+**Warning: this will delete your development database.**
```
withinfocus, line 254: There's a `warn` decorator you can use here, but not sure if it works within a surrounding block.

withinfocus (commented):
Only jumped around for a moment and not a full review, but I think you have link path issues it can't resolve so the build's failing.

Once this renders and we can see a preview it'll be easier to read.

docs/getting-started/server/guide-vscode.md
tangowithfoxtrot: I updated the language and formatting for this process to hopefully make it more clear.

withinfocus:
@tangowithfoxtrot my other comment may have been missed -- link errors are blocking the build.

tangowithfoxtrot:
> ~as it's a direct copy from the existing "Setup Guide" doc and works with my local build.~

Whoops. That was incorrect 😁 Looking into it now...

docs/getting-started/server/guide-vscode.md
```diff
@@ -0,0 +1,270 @@
...
+
+   ```bash
+   git config --local core.hooksPath .git-hooks
+   ```
+
+   Formatting requires a full build, which may be too slow to do every commit. As an alternative,
+   you can run `dotnet format` from the command line when convenient (e.g. before requesting a PR
+   review).
+
+## Configure Env File
```
withinfocus, line 62: 💭 Can we start this documentation right here? The prior two steps are a repeat of the main setup guide.

docs/getting-started/server/database/_category_.yml
```diff
@@ -1,2 +1,2 @@
 label: "Databases"
-position: 2
+position: 3
```
withinfocus, line 2: ⚠️ As I read through all this I started to feel like the guide in this PR is a sub-page of the server's getting started page since there's repeated content. I think this needs some organizing and you wouldn't need this move.

withinfocus (changes requested):
I think it'd be a good idea to pull the repeated content and see how we can thin this new documentation down.

I am definitely interested in using this!

docs/getting-started/server/guide-vscode.md
```diff
@@ -0,0 +1,270 @@
...
+- Latin lowercase letters (a through z)
+- Base 10 digits (0 through 9)
+- Non-alphanumeric characters such as: exclamation point (!), dollar sign ($), number sign (#), or
+  percent (%).
+
+:::
+
+4.  You can change the other variables or use their default values. Save and quit this file.
+
+### SQL Server
```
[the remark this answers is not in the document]
tangowithfoxtrot: The first thought that comes to mind if we wanted to have slight variance in the steps that _do_ still manually require some user interaction would be to add a "Dev Container" option to the top-right dropdown menu, or something like that.

Doing so might also be weird though because that menu currently distinguishes between internal engineering and external community dev configuration guides.

docs/getting-started/server/database/_category_.yml
Hinton: I agree, maintaining the cloud and self-hosted configs are already confusing enough and it would be nice to have a single guide that gives you two options towards the end.

tangowithfoxtrot: I'd love to integrate the container setup into the standard setup guide, but I am not sure how to do so, as even the sections that are shared between the two setups have a few steps that vary slightly.

Is there some way to indicate variance of the setup process _at the step level_ without adding an entire "Dev Container - Bitwarden Dev" "Dev Container - Community Dev" option in the dropdown menu?

Hinton: Maybe [tabs](https://docusaurus.io/docs/markdown-features/tabs)?

withinfocus:
With https://github.com/bitwarden/server/pull/3080 landed should we look at this again?

withinfocus:
I'd suggest just committing to that older PR and to apply my direction on the code vs. contributing docs.

PR #155, opened 2023-07-13
ADR on observability

withinfocus:
## Objective

ADR on embracing OpenTelemetry. Relates to #156. Successor of #151.

### DIFF
--- a/docs/architecture/adr/0020-observability-with-opentelemetry.md
+++ b/docs/architecture/adr/0020-observability-with-opentelemetry.md
@@ -14,9 +14,9 @@ tags: [server]
 Along with the maturation of the codebase over the years, the userbase of the platform has also
 grown significantly and more insight is needed into how services are performing at a fine-grained
 level. External profilers can certainly be attached in any running environment, but the platform
-itself needs to offer internal metrics not just to support customers running the product but to
-enable engineers to improve it and tackle performance issues with solid data and evidence as to what
-and why something should change.
+itself needs to offer internal metrics not just to support self-hosted customers running the product
+but to enable engineers to improve it and tackle performance issues with solid data and evidence as
+to what and why something should change.
 
 ## Considered Options
 
@@ -50,6 +50,8 @@ Chosen option: **Use open observability standards**.
 
 - Proprietary profiler implementations may offer signal information that OpenTelemetry can't,
   including automatic instrumentation.
+- With the capability to capture signals within the platform comes the burden of needing to maintain
+  clear policies around not capturing sensitive data.
 
 ### Plan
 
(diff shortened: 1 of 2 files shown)


justindbaur:
Another possibility could be migrating from the [DD Profiler](https://docs.datadoghq.com/profiler/enabling/dotnet/?tab=linux) we use now to the [Open Telemetry](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation) one while still using the DD agent as the collector. This way we still just rely on the CORECLR profiler to send out the metrics when the infrastructure has been setup. 

The benefits would be, 
- It's incredibly similar to how we collect metrics now, which means there would be less chance for our metrics collection to cause new performance degradation.
- No new dependencies in source code
- Switching to the OTEL profiler should still give us custom metrics (I have asked DD if adding custom metrics support would be possible)
- No extra dependencies for the 99% of SH users who won't ever setup metrics exporting
- SH users who do want to collect metrics aren't locked into OTEL, most of the big services offer a profiler that could be setup. [ref](https://grep.app/search?q=CORECLR_PROFILER%3D%7B)
  - Technically, SH users can already do this today and adding OTEL SDK doesn't lock them into it but if we went the profiler route we would probably document it making it a bit more official, yet still for the advanced users.

The downside is:
- The profilers are a more advanced to configure
  - SH users could still use them and can today but would require them to make a custom image (example below)

```Dockerfile
# Based of our supplied image
FROM bitwarden/self-host

# Install the profiler and set it up
ARG OTEL_VERSION=0.7.0
ADD https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/download/v${OTEL_VERSION}/otel-dotnet-auto-install.sh otel-dotnet-auto-install.sh
RUN apt-get update && apt-get install -y unzip && \
    OTEL_DOTNET_AUTO_HOME="/otel-dotnet-auto" sh otel-dotnet-auto-install.sh
```

withinfocus:
For Bitwarden at least, our new presence in k8s has the latest DD agent installed. My intention was for us to then configure the agent such that it exposes a collector for OTLP, documented at https://docs.datadoghq.com/opentelemetry/otlp_ingest_in_the_agent/?tab=host. I am concerned that the OpenTelemetry profiler isn't quite production-ready yet and we know how to set the DD agent up for logs already.

The biggest benefit that I wanted is our ability to openly collect custom metrics via code changes. I realize that doesn't likely benefit a SH setup, but we do need it for ourselves and I felt it was a valuable feature addition. Do you think we can get by with https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation and staying entirely native? Is there really a performance risk to using OpenTelemetry libraries in code?

justindbaur:
Pretty much no matter what I would use the DD agent as the collector. This is just about how exactly we push the traces/metrics from our application. 

I just wanted to throw out a way to do it that is more like how we do it now, plus now getting custom metrics. I do not know if there would be any performance hit, both the profilers use an architecture like this: 
OTEL Architecture

So with the SDK in our Application IL, we have to devote a thread to listening and pushing the metrics and with the profiler the CLR does that for us. BUT the CLR is still on the same machine as our application so I can't imagine the difference is huge but wouldn't be surprised if there is _a_ difference.

docs/architecture/adr/0020-observability-with-opentelemetry.md
```diff
@@ -0,0 +1,121 @@
...
+ability to specify a gRPC or HTTP endpoint for OTLP. Segmentation of activities will continue to be
+made using the configurable `ProjectName`.
+
+The initial implementation will provide default instrumentation details coming from ASP.NET Core and
+any used HTTP clients. Within Bitwarden the automatic instrumentation (profiler) may be explored at
+a future date but a code-first solution is desired to allow for more control and less setup during
+installation. It is expected that local processes will ingest logs / exports as desired.
+
+Software development lifecycle enhancements will be made to clarify best practices and review
+requirements for logging or monitoring changes. A [deep dive][../../deep-dives] will be added on
```
Hinton, line 102: Link is broken

Hinton (approved):

MGibson1 (approved):

PR #158, opened 2023-07-16
Portal naming changes

trmartin4:
## Objective

The term "Bitwarden Portal" only refers to the cloud-hosted portal that we use internally at Bitwarden to administer the cloud-hosted instance.  For self-hosted customers, the same app is referred to as the "System Management Portal" in our customer-facing documentation and help system.

I've changed the menu option to "System Management Portal" and also added a Bitwarden-specific note about the different naming.

Unrelated, I moved the `Setup` section ahead of the `Configuring Access` section because it flows more with how a user would approach the application.

### DIFF
--- a/docs/getting-started/server/portal.md
+++ b/docs/getting-started/server/portal.md
@@ -1,12 +1,57 @@
-# Bitwarden Portal
+# System Management Portal
+
+:::info Portal naming
+
+This documentation refers to the deployment of the `Admin` application in our `server` repository.
+To disambiguate this application from others in the Bitwarden landscape, we refer to it as follows:
+
+- For **Cloud-Hosted** Instances (internal to Bitwarden) &rarr; Bitwarden Portal
+- For **Self-Hosted** Instances &rarr; System Management Portal
+
+:::
+
+## Setup
+
+1.  Navigate to the `server/src/admin` directory.
+2.  Restore nuget packages:
+
+    ```bash
+    dotnet restore
+    ```
+
+3.  Install npm packages:
+
+    ```bash
+    npm ci
+    ```
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/server/portal.md
```diff
@@ -1,12 +1,61 @@
...
+
+This documentation refers to the deployment of the `Admin` application in our `server` repository.
+To disambiguate this application from others in the Bitwarden landscape, we refer to it as follows:
+
+- For **Cloud-Hosted** Instances (internal to Bitwarden) &rarr; **Bitwarden Portal**
+- For **Self-Hosted** Instances &rarr; **System Management Portal**
+
+:::
+
+</bitwarden>
```
eliykat, line 15: This is an important (and potentially confusing) naming convention, I tend to think it should be shown to everyone (i.e. not wrapped in `<bitwarden>` tags).

docs/getting-started/server/portal.md
```diff
@@ -1,12 +1,61 @@
-# Bitwarden Portal
+# System Management Portal
+
+<bitwarden>
+
+:::info Portal naming
+
+This documentation refers to the deployment of the `Admin` application in our `server` repository.
```
eliykat, line 7: Nit: I think these are referred to as projects?

```suggestion
This documentation refers to the deployment of the `Admin` project in our `server` repository.
```

eliykat (changes requested):

docs/getting-started/server/portal.md
trmartin4: Good idea.  I've removed the tags.

eliykat (approved):

PR #160, opened 2023-07-17
Add Yubikey 2FA docs and other server-side docs cleanup

trmartin4:
## Objective

- Added documentation to the Advanced Server Setup for how to set up YubiKey 2FA in local development.  This came up on the Auth team with some recent tickets and I had to look in Slack history for recommendations.
- Added `sidebar_position` tags to all of the menu items for the Server setup so that they are alphabetized by title (after the setup guide, which is first)
- Moved User Secrets Reference from Getting Started to Contributing, as this felt like documentation that would be used while developing the code ("contributing") instead of initial setup where we have the helper script defined.

### DIFF
--- a/docs/contributing/user-secrets.md
+++ b/docs/contributing/user-secrets.md
@@ -2,10 +2,10 @@
 
 ## Manually editing user secrets
 
-We recommend using the automated helper script described in the [Server Setup Guide](./guide.md).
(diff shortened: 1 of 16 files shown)


jlf0dev (approved):
Thanks for taking the time to do this!

PR #169, opened 2023-08-07
Fix feature flags user secrets key and minor edits

eliykat:
## Objective

Update the feature flags page to:
* clarify that the `launchDarkly::flagValues` keys need to be nested under `globalSettings`
* refer to `secrets.json` rather than AppSettings, as that is our recommended workflow
* improve the structure of this section


docs/contributing/feature-flags.md
```diff
@@ -80,15 +84,12 @@ To set up a data source via a local file, create one with the following:
...
 ```
 
-By default, the LaunchDarkly startup will look for this file to be named `flags.json` and be placed
-in the root project directory (e.g. `/src/Api/` for the `Api` project), where it will be deployed to
-the build output directory. However, if you prefer to store the file in a different location, the
-`FlagDataFilePath` configuration setting can be used to override it. The file must be present before
-building the solution, but once there you can change the file contents and see immediate results in
-running / debugging code.
-
-In either case replace `example-boolean-key` and `example-string-key` with your flag names and
```
withinfocus, line 90: ❓ Did you feel this was unnecessary? Per my other comment this line could be moved above the sections too if you wanted.

withinfocus (changes requested):
Organizational request.

docs/contributing/feature-flags.md
```diff
@@ -43,33 +43,37 @@ applies to self-hosted.
 
 :::
 
-### Local configuration
+### Local configuration - user secrets
 
-As shown above, local server development instances will not query LaunchDarkly for feature flag
-values.
+Local development server instances will not query LaunchDarkly for feature flag values.
```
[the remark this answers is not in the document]
eliykat: Done

docs/contributing/feature-flags.md
eliykat: I thought the sample values were kind of self-explanatory, but I don't mind adding this back in to avoid any confusion. However, I think it only makes sense in the context of the example JSON, so I've just duplicated in each section. (Some minimal duplication for clarity seems fine to me.)

withinfocus (approved):
Good additions!

PR #194, opened 2023-09-13
setup_secrets: Pass `-clear` as switch

amorask-bitwarden:
## Objective

While going through the onboarding, I had to reset my user secret several times. To do so, I was trying to use the `-clear` parameter as specified in the docs: `pwsh setup_secrets.ps1 -clear:$True`

However, I noticed [that this part of the script](https://github.com/bitwarden/server/blob/master/dev/setup_secrets.ps1#L15) was never being hit.

I changed `$True` to `true` and it started to work:

image

### DIFF
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -363,7 +363,7 @@ repository.
     pwsh setup_secrets.ps1
     ```
 
-The helper script also supports an optional `-clear` flag which removes all existing settings before
+The helper script also supports an optional `-clear` switch which removes all existing settings before
 re-applying them:
 
 ```bash
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -363,8 +363,8 @@ repository.
     pwsh setup_secrets.ps1
     ```
 
-The helper script also supports an optional `-clear` switch which removes all existing settings before
-re-applying them:
+The helper script also supports an optional `-clear` switch which removes all existing settings
+before re-applying them:
 
 ```bash
 pwsh setup_secrets.ps1 -clear
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -363,11 +363,11 @@ repository.
     pwsh setup_secrets.ps1
     ```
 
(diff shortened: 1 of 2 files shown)


withinfocus:
Is it the casing? https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_booleans suggests lowercase T.

amorask-bitwarden:
> Is it the casing? https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_booleans suggests lowercase T.

@withinfocus I thought that too, but I couldn't trigger the clear with `$true` either:
image

Hinton (approved):

Hinton:
What happens if you try `$true`, `$false` in a powershell terminal?

image

amorask-bitwarden:
> What happens if you try `$true`, `$false` in a powershell terminal?
> 
> image

@Hinton That works:
image

So my guess is that when passing the parameters in from a unix shell to the `pwsh` executable, you have to use `true`, but if you're already in a PowerShell context, it's `$True`. Do you think it's worth splitting the documentation at that point for Mac/Linux vs. Windows?

Hinton:
I switched the bool to switch which removes the need to include `:$True`. https://github.com/bitwarden/server/pull/3269

amorask-bitwarden:
> I switched the bool to switch which removes the need to include `:$True`. [bitwarden/server#3269](https://github.com/bitwarden/server/pull/3269) Can you update the docs to note that it's a switch and doesn't need any value?

@Hinton Updated in the rebased commit: https://github.com/bitwarden/contributing-docs/pull/194/commits/4e4a299bfabff09a3a0286e900a9b6c06f843ae1

docs/getting-started/server/guide.md
```diff
@@ -363,11 +363,11 @@ repository.
     pwsh setup_secrets.ps1
     ```
 
-The helper script also supports an optional flag which removes all existing settings before
+The helper script also supports an optional `-clear` flag which removes all existing settings before
```
Hinton, line 366:
```suggestion
The helper script also supports an optional `-clear` switch which removes all existing settings before
```

PR #210, opened 2023-10-06
Mark ADR-0006 as done

Hinton:


withinfocus (approved):
For when the dependency is merged.

PR #215, opened 2023-10-18
[PM-4407] docs: add cd command to setup instructions

anatelli10:
Update clients setup instructions

Add the current directory command to the "Install the dependencies" setup instructions step.

## Objective

This would make it so all necessary commands are included in the doc. And, by adding it into an existing step, brevity is maintained.


eliykat (approved):
Thanks for the contribution!

PR #244, opened 2023-12-01
Don't recommend closing PRs

withinfocus:
## Objective

Don't recommend to close PRs if the intention is to later address them. Renovate will not re-prompt to re-create a PR if you close it and while that may be fine when dealing with minors and patches, closing major version upgrades could be considered a long-term if not permanent closure and we'd lose oversight with the dependency.

I thought about expanding on this more to define context or major vs. minor, etc. but that could be even more confusing.


Hinton:
The teams are aware that if they close a PR they are responsible for monitoring it and re-visiting it in the future. I don't think it's viable to have PRs open for months which is the timelines we usually delay .NET and Angular migrations.

withinfocus:
Adjusted the language a bit to make your point clear. In some repos I think it's acceptable to leave a PR open, and others it isn't; what I don't want though is someone closing one and not knowing that offline monitoring is expected.

Hinton (approved):
Looks good, thanks.

PR #248, opened 2023-12-05
Add State Provider Framework Deep Dive

justindbaur:
## Objective

To document the newly added State Provider Framework added in bitwarden/clients#6640 and refined in bitwarden/clients#6860.


docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,308 @@
...
+:::
+
+### Example PRs
+
+_TODO: Include PR's_
+
+## Testing
+
+Testing business logic with data and observables can sometimes be cumbersome, to help make that a
+little easier, there are a sweet of helpful "fakes" that can be used instead of traditional "mocks".
```
MGibson1, line 246:
```suggestion
little easier, there are a suite of helpful "fakes" that can be used instead of traditional "mocks".
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,308 @@
...
+
+_TODO: Refine user story_
+
+## Advanced Usage
+
+### `update`
+
+It was mentioned earlier that update has the type signature
+`update(updateState: (state: T) => T): Promise<T>` but that is not entirely true. You can treat it
+as if it has that signature and many will but it's true signature is the following:
```
MGibson1, line 258:
```suggestion
The update method has options defined as follows:
```

docs/architecture/deep-dives/state/index.md
justindbaur: 🤦

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,308 @@
...
+  }
+}
+```
+
+:::note
+
+`getAccounts` only gets data from the legacy account object that was used in `StateService`. As data
+gets migrated off of that account object the response from `getAccounts`, which returns a record
+where the key will be a users id and the value being the legacy account object. It can continue to
+be used to retrieve the known authenticated user ids though.
```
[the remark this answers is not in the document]
justindbaur: Should we make that type of method already or wait until there is need to make it?

withinfocus (commented):
Seagulling with just a couple things I noticed with a quick read, really for my OCD.

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,379 @@
...
+
+`KeyDefinition` builds on the idea of [`StateDefinition`](#statedefinition) but it gets more
+specific about the data to be stored. `KeyDefinition`s can also be instantiated in your own teams
+code. This might mean creating it in the same file as the service you plan to consume it or you may
+want to have a single `key-definitions.ts` file that contains all the entries for your team. Some
+example instantiations are:
+
+```typescript
+const MY_DOMAIN_DATA = new KeyDefinition<MyState>(MY_DOMAIN_DISK, "data", {
+  deserializer: (jsonData) => null, // convert to your data from json
```
audrey-jensen, line 78: 💭 I found this a bit misleading. `jsonData` is already parsed, whereas I expect JSON to be the stringified version of data. Had there been an implementation provided or there were a type annotation or something, I'd not have been as confused.

Example:
```typescript
const MY_DOMAIN_DATA = new KeyDefinition<MyState>(MY_DOMAIN_DISK, "data", {
  // convert to your data from serialized representation `{ foo: string }` 
  // to a fully-typed `MyState`.
  deserializer: (jsonData) => MyState.fromJson(jsonData), 
});
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,379 @@
...
+```
+
+The `shouldUpdate` option can be useful to help avoid an unnecessary update, and therefore avoid an
+unnecessary emission of `state$`. You might want to use this to avoid setting state to `null` when
+it is already `null`. The `shouldUpdate` method gives you in it's first parameter the value of state
+before any change has been made to it and the dependency you have, optionally, provided through
+`combineLatestWith`. To avoid setting `null` twice you could call `update` like below:
+
+```typescript
+await myUserState.update(() => null, { shouldUpdate: (state) => state != null });
```
audrey-jensen, line 316: ❓ Is this accurate? Based on the description I'd expect the code to never update state once it's null. That is, if I manage to set `state === null`, then `shouldUpdate` would always return false, and thus cancel **every** update.

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,379 @@
...
+`data => new Date(data)`.
+
+`cleanupDelayMs` (optional) - Takes a number of milliseconds to wait before cleaning up the state
+after the last subscriber has unsubscribed. Defaults to 1000ms.
+
+### `StateProvider`
+
+`StateProvider` is an injectable service that includes 3 methods for getting state. These three
+methods are helpers for invoking their more modular siblings `ActiveStateProvider.get`,
+`SingleUserStateProvider.get`, and `GlobalStateProvider.get` and they have the following type
```
audrey-jensen, line 123: ❓ Is it ok to use the more tightly scoped siblings? Like, if I only care about the active user, can I just inject `ActiveUserStateProvider`?

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,379 @@
...
+## Testing
+
+Testing business logic with data and observables can sometimes be cumbersome, to help make that a
+little easier, there are a suite of helpful "fakes" that can be used instead of traditional "mocks".
+Now instead of calling `mock<StateProvider>()` into your service you can instead use
+`new FakeStateProvider()`.
+
+_TODO: Refine user story_
+
+## Advanced usage
```
audrey-jensen, line 290: 🤔 How do I manage state that is encrypted on disk but plaintext in memory? I imagine I'd want to use a disk and a memory store and wire them together in sequence?

1. RxJs application sequence updates memory store.
2. Memory store triggers Rxjs encryption sequence.
3. Rxjs encryption sequence updates disk store.

All of the back-and-forth with application logic would end up going back and forth with the memory store. The disk store would be used for loading, then disconnect. Later, when an update comes through it'd load long enough to save the update then go quiet again. I imagine that means that it's timeout could be kept pretty short--perhaps on the order of milliseconds, or perhaps even immediately?

docs/architecture/deep-dives/state/index.md
justindbaur: You can, the siblings are all injectable through DI or available right next to `StateProvider` (browser background or CLI). These providers all have a caching element to them where if state is requested for a `KeyDefinition` more than once they will make it the first time and return the same instance for any subsequent requests. But that cache is inside the siblings themselves so you won't even be skirting that by using `ActiveUserStateProvider` directly.

docs/architecture/deep-dives/state/index.md
justindbaur: We are working on this exact problem! @MGibson1 has a draft PR right now where we are working on a good solution to these types of problems. bitwarden/clients#7290. So far everything we've thrown out is very inline with what you've mentioned here. Except for one thing, I think right now the decrypted state is going to be read-only. If you want to update it we are going to make you manually re-encrypt it and then call `update` to your disk state. But I've mentioned Matt above so he can maybe take this for some ideas!

docs/architecture/deep-dives/state/index.md
audrey-jensen: 🎨 Sweet! I might make it more explicit that if you only need one kind of state, that you should inject the smaller interface.

docs/architecture/deep-dives/state/index.md
audrey-jensen: Oh, I always figured the encryption and decryption was manually performed, since there's no way the storage layer could differentiate.

Now that I think about it, though, there's some interesting possibilities that come to mind if you were to divide the state into encrypted parts and plaintext parts. You could sync the encrypted parts using content-addressable storage and k-anonymity tokens stored with the plaintext. That's probably way out of scope of this, though.

docs/architecture/deep-dives/state/index.md
justindbaur: Very, very fair. I took this exact snippet and committed it, thanks!

docs/architecture/deep-dives/state/index.md
justindbaur: Yep, good call. I Reworded the paragraph above to read:

> The `shouldUpdate` option can be useful to help avoid an unnecessary update, and therefore avoid an
unnecessary emission of `state$`. You might want to use this to avoid setting state to `null` when
it is already `null`. The `shouldUpdate` method gives you in it's first parameter the value of state
before any change has been made to it and the dependency you have, optionally, provided through
`combineLatestWith`. To avoid setting `null` to your state when it's already `null` you could call
`update` like below:

Do you think that is clearer?

docs/architecture/deep-dives/state/index.md
justindbaur: I tried to make it more explicit as an option but I didn't go as far as to say they _should_ do it that way. It has a con where they will need to inject another service if they now have to go change the code that instantiates that service a minimum of 4 other places. Heaviness-wise the heaviest service is going to be `ActiveUserStateProvider` which is likely to be the most commonly used service. All the other providers take the same dependencies minus `AccountService`. So if they can use `ActiveUserStateProvider` they will get no new circular dependencies possibilities from just taking the `StateProvider` helper.

docs/architecture/deep-dives/state/index.md
justindbaur: The decryption _is_ still performed by code written by each owner of that state, you just have to give us that function and we decide _when_ to run it. Basically we will run it for you when you have updated the encrypted data AND someone is subscribed to the derived data (someone cares ❤️). 

We didn't do anything near as fancy as you described and we have merged the derived state PR but if you want to have a chat about the pros of doing it like you described I'd be down to listen. As of now we do have to pass around user key access and crypto services every pretty much which I am definitely not in love with.

MGibson1 (commented):
I'm happy with this, but I'll leave it to @withinfocus since I worked on this quite a bit.

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+now, don't migrate data that is stored in secure storage but please contact platform when you have
+data you wanted to migrate so we can prioritize a long term solution. If you need new data in secure
+storage, use `StateService` for now.
+
+:::
+
+#### `KeyDefinition`
+
+`KeyDefinition` builds on the idea of [`StateDefinition`](#statedefinition) but it gets more
+specific about the data to be stored. `KeyDefinition`s can also be instantiated in your own teams
```
withinfocus, line 71:
```suggestion
specific about the data to be stored. `KeyDefinition`s can also be instantiated in your own team's
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+want to have a single `key-definitions.ts` file that contains all the entries for your team. Some
+example instantiations are:
+
+```typescript
+const MY_DOMAIN_DATA = new KeyDefinition<MyState>(MY_DOMAIN_DISK, "data", {
+  // convert to your data from serialized representation `{ foo: string }` to fully-typed `MyState`
+  deserializer: (jsonData) => MyState.fromJSON(jsonData),
+});
+
+// Or if your state is an array, use the built in helper
```
withinfocus, line 82:
```suggestion
// Or if your state is an array, use the built-in helper
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+  "data",
+  {
+    deserializer: (jsonDataElement) => MyState.fromJSON(jsonDataElement), // provide a deserializer just for the element of the array
+  },
+);
+
+// record
+const MY_DOMAIN_DATA: KeyDefinition<Record<string, MyStateElement>> =
+  KeyDefinition.record<MyStateValue>(MY_DOMAIN_DISK, "data", {
+    deserializer: (jsonDataValue) => MyState.fromJSON(jsonDataValue), // provide a deserializer just for the value in each key-valu pair
```
withinfocus, line 94:
```suggestion
    deserializer: (jsonDataValue) => MyState.fromJSON(jsonDataValue), // provide a deserializer just for the value in each key-value pair
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+  });
+```
+
+The first argument to `KeyDefinition` is always the `StateDefinition` that this key should belong
+to. The second argument should be a human readable, camelCase formatted name of the `KeyDefinition`.
```
withinfocus, line 99:
```suggestion
to. The second argument should be a human readable, camelCase-formatted name of the `KeyDefinition`.
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+`KeyDefinition`s that consume the same `StateDefinition`. The responsibility of this uniqueness is
+on the team. As such, you should never consume the `StateDefinition` of another team in your own
+`KeyDefinition`. The third argument is an object of type
+[`KeyDefinitionOptions`](#keydefinitionoptions).
+
+##### `KeyDefinitionOptions`
+
+`deserializer` (required) - Takes a method that gives you your state in it's JSON format and makes
+you responsible for converting that into JSON back into a full JavaScript object, if you choose to
+use a class to represent your state that means having it's prototype and any method you declare on
```
withinfocus, line 111:
```suggestion
use a class to represent your state that means having its prototype and any method you declare on
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+`data => new Date(data)`.
+
+`cleanupDelayMs` (optional) - Takes a number of milliseconds to wait before cleaning up the state
+after the last subscriber has unsubscribed. Defaults to 1000ms.
+
+### `StateProvider`
+
+`StateProvider` is an injectable service that includes 4 methods for getting state. These four
+methods are helpers for invoking their more modular siblings `ActiveStateProvider.get`,
+`SingleUserStateProvider.get`, `GlobalStateProvider.get`, and `DerivedStateProvider`, these siblings
```
withinfocus, line 124:
```suggestion
`SingleUserStateProvider.get`, `GlobalStateProvider.get`, and `DerivedStateProvider`. These siblings
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+  getGlobal<T>(keyDefinition: KeyDefinition<T>): GlobalState<T>;
+  getDerived<TFrom, TTo, TDeps>(
+    parentState$: Observable<TFrom>,
+    deriveDefinition: DeriveDefinition<TFrom, TTo, TDeps>,
+    dependenciess: TDeps,
+  );
+}
+```
+
+A very common practice will be to inject `StateProvider` in your services constructor and call
```
withinfocus, line 142:
```suggestion
A very common practice will be to inject `StateProvider` in your service's constructor and call
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+`getActive`, `getGlobal`, or both in your constructor and then store private properties for the
+resulting `ActiveUserState<T>` and/or `GlobalState<T>`. It's less common to need to call `getUser`
```
withinfocus, line 144:
```suggestion
resulting `ActiveUserState<T>` and / or `GlobalState<T>`. It's less common to need to call `getUser`
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+  async clear(userId: UserId): Promise<void> {
+    await this.stateProvider.getUser(userId, FOLDERS_USER_STATE).update((state) => null);
+  }
+}
+```
+
+### `ActiveUserState<T>`
+
+`ActiveUserState<T>` is an object to help you maintain and view the state of the currently active
+user. If the currently active user changes, like through account switching. The data this object
```
withinfocus, line 176:
```suggestion
user. If the currently active user changes, like through account switching, the data this object
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+- The active user changes.
+- The chosen storage location emits an update to the key defined by `KeyDefinition`. This can occur
+  for any reason including:
+  - You caused an update through the `update` method.
+  - Another service in a different context calls `update` on their own instance of
+    `ActiveUserState<T>` made from the same `KeyDefinition`.
+  - A `SingleUserState<T>` method pointing at the same `KeyDefinition` as `ActiveUserState` and
+    pointing at the user that is active that had `update` called
+  - Someone updates the key directly on the underlying storage service (Please don't do this)
```
withinfocus, line 213:
```suggestion
  - Someone updates the key directly on the underlying storage service _(please don't do this)_
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+### `GlobalState<T>`
+
+`GlobalState<T>` has an incredibly similar API surface as `ActiveUserState<T>` except it targets
+global scoped storage and does not emit an update to `state$` when the active user changes, only
```
withinfocus, line 218:
```suggestion
global-scoped storage and does not emit an update to `state$` when the active user changes, only
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+    await Promise.all(accounts.map(({ userId, account }) => updateAccount(userId, account)));
+  }
+}
+```
+
+:::note
+
+`getAccounts` only gets data from the legacy account object that was used in `StateService`. As data
+gets migrated off of that account object the response from `getAccounts`, which returns a record
+where the key will be a users id and the value being the legacy account object.
```
withinfocus, line 279:
```suggestion
where the key will be a user's ID and the value being the legacy account object.
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+_TODO: Implement method just for user ids_
```
withinfocus, line 281: ❓ More?

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+:::
+
+### Example PRs
+
+_TODO: Include PR's_
```
withinfocus, line 287: ❓ And more?

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+## Testing
+
+Testing business logic with data and observables can sometimes be cumbersome, to help make that a
```
withinfocus, line 291:
```suggestion
Testing business logic with data and observables can sometimes be cumbersome. To help make that a
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+little easier, there are a suite of helpful "fakes" that can be used instead of traditional "mocks".
```
withinfocus, line 292:
```suggestion
little easier there are a suite of helpful "fakes" that can be used instead of traditional "mocks".
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+Now instead of calling `mock<StateProvider>()` into your service you can instead use
+`new FakeStateProvider()`.
+
+_TODO: Refine user story_
```
withinfocus, line 296: ❓ And more?

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+type StateUpdateOptions = {
+  shouldUpdate?: (state: T, dependency: TCombine) => boolean;
+  combineLatestWith?: Observable<TCombine>;
+  msTimeout?: number
+}
+```
+
+The `shouldUpdate` option can be useful to help avoid an unnecessary update, and therefore avoid an
+unnecessary emission of `state$`. You might want to use this to avoid setting state to `null` when
+it is already `null`. The `shouldUpdate` method gives you in it's first parameter the value of state
```
withinfocus, line 319:
```suggestion
it is already `null`. The `shouldUpdate` method gives you in its first parameter the value of state
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+`update` like below:
+
+```typescript
+await myUserState.update(() => null, { shouldUpdate: (state) => state != null });
+```
+
+The `combineLatestWith` option can be useful when updates to your state depend on the data from
+another stream of data. In
+[this example](https://github.com/bitwarden/clients/blob/2eebf890b5b1cfbf5cb7d1395ed921897d0417fd/libs/common/src/auth/services/account.service.ts#L88-L107)
+you can see how we don't want to set a user id to the active account id unless that user id exists
```
withinfocus, line 331:
```suggestion
you can see how we don't want to set a user ID to the active account ID unless that user ID exists
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+  throw new Error();
+}
+await this.activeAccountIdState.update(() => userId);
+```
+
+The use of the `combineLatestWith` option is preferred because it fixes a couple subtle issues.
+First, the use of `firstValueFrom` with no `timeout`. Behind the scenes we enforce that the
+observable given to `combineLatestWith` will emit a value in a timely manner, in this case a
+`1000ms` timeout but that number is configurable through the `msTimeout` option. The second issue it
+fixes, is that we don't guarantee that your `updateState` function is called the instant that the
```
withinfocus, line 346:
```suggestion
fixes is that we don't guarantee that your `updateState` function is called the instant that the
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+`update` method is called. We do however promise that it will be called before the returned promise
+resolves or rejects. This may be because we have a lock on the current storage key. No such locking
+mechanism exists today but it may be implemented in the future. As such, it is safer to use
+`combineLatestWith` because the data is more likely to retrieved closer to when it needs to be
+evaluated.
+
+## FAQ
+
+### Do I need to have my own in memory cache?
```
withinfocus, line 355:
```suggestion
### Do I need to have my own in-memory cache?
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+If you previously had a memory cache that exactly represented the data you stored on disk (not
+decrypted for example), then you likely don't need that anymore. All the `*State` classes maintain
+an in memory cache of the last known value in state for as long as someone is subscribed to the
+data. The cache is cleared after 1000ms of no one subscribing to the state though. If you know you
+have sporadic subscribers and a high cost of going to disk you may increase that time using the
+`cleanupDelayMs` on `KeyDefinitionOptions`.
+
+### I store my data as a Record/Map but expose it as an array, what should I do?
```
withinfocus, line 364:
```suggestion
### I store my data as a Record / Map but expose it as an array -- what should I do?
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+Give `KeyDefinition<T>` generic the record shape you want, or even use the static `record` helper
+method. Then to convert that to an array that you expose just do a simple
+`.pipe(map(data => this.transform(data)))` to convert that to the array you want to expose.
+
+### Why KeyDefinitionLike
+
+`KeyDefinitionLike` exists to help you create a frozen in time version of your `KeyDefinition` this
```
withinfocus, line 372:
```suggestion
`KeyDefinitionLike` exists to help you create a frozen-in-time version of your `KeyDefinition`. This
```

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+is helpful in state migrations so that you don't have to import something from the greater
+application which is something that should rarely happen.
+
+### When does my deserializer run?
+
+The `deserialier` that you provide in the `KeyDefinitionOptions` is used whenever your state is
+retrieved from a storage service that stores it's data as JSON. All disk storage services serialize
```
withinfocus, line 379:
```suggestion
retrieved from a storage service that stores its data as JSON. All disk storage services serialize
```

withinfocus (commented):
I started getting a bit wild with copyediting. There are a few still-missing sections I assume need filling in, or strike them from the PR perhaps?

docs/architecture/deep-dives/state/index.md
audrey-jensen: Yeah, it's more clear. I still think there's an issue, though. As written, the example will never send a request when its value is presently null. I think the sample would need to be adjusted to compare vs. some closed over variable to reflect what the paragraph is saying.

```typescript
const proposedValue = // ...
await myState.update(() => proposedvalue, { shouldUpdate: (currentValue) => currentValue !== null || proposedValue !== null });
```

Is that right?

docs/architecture/deep-dives/state/index.md
audrey-jensen: I think we kind of did have this conversation, though without the CAS portions. If I have a chance to experiment with it as part of Sends, then I'll let you know about the branch so we can figure out if it's something that we should polish and lift.

docs/architecture/deep-dives/state/index.md
audrey-jensen: Spoke out-of-band about this. The missing context is **why** you'd want to set it to null. Since the code above categorically sets the value to null, as might occur in `clear()`, checking the current state is all that's needed. The sample I provided is also correct, but it introduces a `proposedValue`, as one might see when writing `save(value)`. Since the logic I'm working with always saves (there is no "delete" or "clear" logic), I'd assumed that context within the code.

audrey-jensen (dismissed):

docs/architecture/deep-dives/state/index.md
```diff
@@ -0,0 +1,388 @@
...
+data into JSON but memory storage differs in this area across platforms. That's why it's imperative
+to include a high quality JSON deserializer even if you think your object will only be stored in
+memory. This can mean you might be able to drop the `*Data` class pattern for your code. Since the
+`*Data` class generally represented the JSON safe version of your state which we now do
+automatically through the `Jsonify<T>` given to your in your `deserializer` method.
+
+## Structure
+
+![State Diagram](State_Diagram.svg)
```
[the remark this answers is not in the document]
trmartin4: I think this probably used Excalidraw, whereas we've previously used PlantUML. I personally like the slightly less formal style of this diagram, but if we'd like standardization I can move this to PlantUML if it would be helpful.

justindbaur:
Well it has only been 8 months but I'd love to try and merge this in 🫤.

withinfocus (dismissed):
Not an exhaustive review because I want to see this merge!

docs/architecture/deep-dives/state/index.md
```diff
@@ -24,21 +24,30 @@ storage framework.
...
+data from the old name to the new name you will lose data. Data pointing at `"memory"` can have its
+name changed.
 
 :::
 
 #### `StateDefinition`
 
+:::note
+
+Secure storage is not currently supported as a storage location in the State Provider Framework. For
```
audreyality, line 38: ❓ Is secure storage just a description, or does it refer to an emerging web api?

docs/architecture/deep-dives/state/index.md
```diff
@@ -47,75 +56,94 @@ export const MY_DOMAIN_DISK = new StateDefinition("myDomain", "disk");
...
-this file and will be looking to make sure that there are no duplicate entries containing the same
-state name and state location. Teams _can_ have the same state name used for both `"disk"` and
-`"memory"` locations. Tests are included to ensure this uniqueness and core naming guidelines so you
-can ensure a review for a new `StateDefinition` entry can be done promptly and with very few
-surprises.
+should be stored.
 
-_TODO: Make tests_
+The Platform team will be responsible to reviewing all new and updated entries in this file and will
+be looking to make sure that there are no duplicate entries containing the same state name and state
```
audreyality, line 62: ⛏️ Consider using present tense.

docs/architecture/deep-dives/state/index.md
```diff
@@ -47,75 +56,94 @@ export const MY_DOMAIN_DISK = new StateDefinition("myDomain", "disk");
...
-`KeyDefinition` builds on the idea of [`StateDefinition`](#statedefinition) but it gets more
-specific about the data to be stored. `KeyDefinition`s can also be instantiated in your own team's
-code. This might mean creating it in the same file as the service you plan to consume it or you may
-want to have a single `key-definitions.ts` file that contains all the entries for your team. Some
-example instantiations are:
+`KeyDefinition` and `UserKeyDefinition` build on the [`StateDefinition`](#statedefinition),
+specifying a single element of state data within the `StateDefinition`.
+
+The framework provides both `KeyDefinition` and `UserKeyDefinition` for teams to use. The
+`UserKeyDefinition` should be used for defining pieces of state that are scoped at a user level.
```
audreyality, line 86: ⛏️ Consider stating this as an instruction. "Use `UserKeyDefinition` for state scoped to a user and `KeyDefinition` for user-independent state." or similar.

docs/architecture/deep-dives/state/index.md
```diff
@@ -47,75 +56,94 @@ export const MY_DOMAIN_DISK = new StateDefinition("myDomain", "disk");
...
-to. The second argument should be a human readable, camelCase-formatted name of the `KeyDefinition`.
-For example, the accounts service may wish to store a known accounts array on disk and choose
-`knownAccounts` to be the second argument. This name should be unique amongst all other
-`KeyDefinition`s that consume the same `StateDefinition`. The responsibility of this uniqueness is
-on the team. As such, you should never consume the `StateDefinition` of another team in your own
-`KeyDefinition`. The third argument is an object of type
-[`KeyDefinitionOptions`](#keydefinitionoptions).
+The arguments for defining a `KeyDefinition` or `UserKeyDefinition` are:
+
+| Argument          | Usage                                                                                                                                                                                             |
```
audreyality, line 126: 👍🏻 Love presenting the options and arguments using tables!

docs/architecture/deep-dives/state/index.md
```diff
@@ -172,6 +200,13 @@ class FolderService {
 
 ### `ActiveUserState<T>`
 
+:::warning
+
+`ActiveUserState` has problems with consider not using it anymore,
```
audreyality, line 205: ⛏️ Looks like you a word. "`ActiveUserState` has race condition problems. Do not use it for updates and consider transitioning your code to SingleUserState instead."

docs/architecture/deep-dives/state/index.md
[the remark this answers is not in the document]
audreyality: Another possibility is to use a markdown-compatible renderer, such as mermaid or GoAT.

audreyality (dismissed):
👍🏻 I also want to see this merged. Feedback is provided, but I'm totally cool with it landing in a followup PR. :)

docs/architecture/deep-dives/state/index.md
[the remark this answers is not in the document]
audreyality: Another possibility is to use a markdown-compatible renderer, such as mermaid or GoAT.

audreyality (approved):

docs/architecture/deep-dives/state/index.md
justindbaur: No emerging API, just an existing API that was consumed in `StateService` that allowed persistent storage of data in a more secure way, the main problem is that is wasn't supported in all clients only in desktop (KeyChain on MacOS for example) and CLI (data was encrypted with the session key). That was one of the reasons it wasn't added in the state provider framework. It's probably deserving of its own document altogether.

audreyality (approved):

PR #258, opened 2023-12-15
Add docs for Public API development

eliykat:
## Objective

Add documentation around
* what the public api is
* general development guidelines
* how to test your changes using SwaggerUI

### DIFF
--- a/docs/getting-started/server/public-api.md
+++ b/docs/getting-started/server/public-api.md
@@ -47,11 +47,11 @@ To use SwaggerUI:
 3. Click "Authorize"
 4. In a separate tab, open the Web Vault and navigate to your Organization Settings page. Click
    "View API Key"
-5. Enter your organization's client_id and client_secret from the Web Vault into Swagger
-6. You can now close the Web Vault and continue in Swagger
-7. In the Scopes section, click "select all"
-8. Click "Authorize" to close the dialog
-9. You should receive a confirmation dialog. Click "Close"
+5. Enter your organization's client_id and client_secret from the Web Vault into Swagger. You can
+   now close the Web Vault and continue in Swagger
+6. In the Scopes section, click "select all"
+7. Click "Authorize" to close the dialog
+8. You should receive a confirmation dialog. Click "Close"
 
 You can now test the Public API by expanding any section, clicking "Try it out", editing the
 request, and clicking Execute. The response will be displayed below. You can also verify that your
--- a/docs/getting-started/server/public-api.md
+++ b/docs/getting-started/server/public-api.md
@@ -1,0 +1,59 @@
+# Public API
+
+The Bitwarden Public API provides organizations a suite of tools for managing members, collections,
+groups, event logs, and policies. More information about the Public API is available in the
+[Help Center](https://bitwarden.com/help/public-api/).
+
+## Differences with the private API
+
+Most developers will be more familiar with the private API used by our client applications. The
+Public API is different in several key areas:
+
+<!-- prettier-ignore -->
+| Private API | Public API |
+| --- | --- |
+| Located at api.bitwarden.com | Located at api.bitwarden.com/public |
+| Used by official Bitwarden client applications | Used by third parties, usually in custom integrations |
+| Broad scope - can be used for anything | Narrow scope - can only be used to manage organizations |
+| Can be changed without notice | Must give notice for any breaking changes
+| Uses user credentials for authentication | Uses organization API key for authentication |
+
+## Development guidelines
+
+1. Avoid making breaking changes - these are any changes that would require existing users of the
+   API to update their integrations to avoid errors or unexpected behavior
+   - for example, make new properties optional, so that existing integrations do not have to supply
+     a value
+2. If you must make breaking changes, consider how to give advance warning to existing users.
+   Communicate with Engineering, Product, and Customer Success Integration teams to coordinate any
+   notice required and minimize impact
+3. Do not use the same request/response models as the private API
+4. Use [xmldoc comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/)
+   to add documentation to your endpoints and models. These will be included in the SwaggerUI output
+
+## Developing locally
+
+When running in dev mode, the Bitwarden server includes a
+[SwaggerUI](https://swagger.io/tools/swagger-ui/) instance, similar to the one found on our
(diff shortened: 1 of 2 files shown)


djsmith85 (dismissed):
Nice addition to the docs

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Differences with the private API
+
+Most developers will be more familiar with the private API used by our client applications. The
+Public API is different in several key areas:
+
+<!-- prettier-ignore -->
+| Private API | Public API |
+| --- | --- |
+| Located at api.bitwarden.com | Located at api.bitwarden.com/public |
```
withinfocus, line 15: 💭 Actually link these?

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+| Used by official Bitwarden client applications | Used by third parties, usually in custom integrations |
+| Broad scope - can be used for anything | Narrow scope - can only be used to manage organizations |
```
withinfocus, line 17:
```suggestion
| Broad scope -- can be used for anything | Narrow scope -- can only be used to manage organizations |
```

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+| Can be changed without notice | Must give notice for any breaking changes
+| Uses user credentials for authentication | Uses organization API key for authentication |
```
withinfocus, line 19:
```suggestion
| Utilizes user credentials for authentication | Utilizes organization API key for authentication |
```

⛏️ Helps break up "use" and "user".

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+Most developers will be more familiar with the private API used by our client applications. The
+Public API is different in several key areas:
+
+<!-- prettier-ignore -->
+| Private API | Public API |
+| --- | --- |
+| Located at api.bitwarden.com | Located at api.bitwarden.com/public |
+| Used by official Bitwarden client applications | Used by third parties, usually in custom integrations |
+| Broad scope - can be used for anything | Narrow scope - can only be used to manage organizations |
+| Can be changed without notice | Must give notice for any breaking changes
```
withinfocus, line 18: ⚠️ I disagree with the phrasing here -- we plan with breaking changes. Perhaps this can just drop the "breaking changes" and the below elaborate on what breaking changes means.

Pointing out https://bitwarden.com/help/bitwarden-software-release-support/ with the notice and windowing could be a benefit too.

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+| Uses user credentials for authentication | Uses organization API key for authentication |
+
+## Development guidelines
+
+1. Avoid making breaking changes - these are any changes that would require existing users of the
```
withinfocus, line 23:
```suggestion
1. Avoid making breaking changes -- these are any changes that would require existing users of the
```

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+2. If you must make breaking changes, consider how to give advance warning to existing users.
+   Communicate with Engineering, Product, and Customer Success Integration teams to coordinate any
+   notice required and minimize impact
+3. Do not use the same request/response models as the private API
+4. Use [xmldoc comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/)
+   to add documentation to your endpoints and models. These will be included in the SwaggerUI output
+
+## Developing locally
+
+When running in dev mode, the Bitwarden server includes a
```
withinfocus, line 36:
```suggestion
When running in dev mode the Bitwarden server includes a
```

⛏️ I find that non-US English content-writers like more comma breakups.

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+[SwaggerUI](https://swagger.io/tools/swagger-ui/) instance, similar to the one found on our
+[Help Center](https://bitwarden.com/help/api/).
+
+SwaggerUI can help you test any changes you make to the Public API without having to write your own
+http requests. You can also check how your xmldoc comments will be presented by SwaggerUI when the
```
withinfocus, line 41:
```suggestion
HTTP requests. You can also check how your xmldoc comments will be presented by SwaggerUI when the
```

docs/getting-started/server/public-api.md
```diff
@@ -0,0 +1,59 @@
...
+
+To use SwaggerUI:
+
+1. Start your local development server (Api and Identity projects) and Web Vault
+2. Navigate to http://localhost:4000/docs
+3. Click "Authorize"
+4. In a separate tab, open the Web Vault and navigate to your Organization Settings page. Click
+   "View API Key"
+5. Enter your organization's client_id and client_secret from the Web Vault into Swagger
+6. You can now close the Web Vault and continue in Swagger
```
withinfocus, line 52: ⛏️ Feels like this can be included in the prior step.

withinfocus (changes requested):
Couple important ones and then some more opinionated.

docs/getting-started/server/public-api.md
eliykat: Sure, although they don't go anywhere

docs/getting-started/server/public-api.md
eliykat: Well this _is_ opinionated

docs/getting-started/server/public-api.md
eliykat: I didn't like this either, good call

docs/getting-started/server/public-api.md
eliykat: By "notice" I meant "notice to third parties/customers". We can and do change the internal api all the time without expressly telling customers. cf. publishing deprecation notices/warnings.

I've tweaked the wording here and included that link. Happy to adjust it further if required, but in that case please provide some suggested wording.

eliykat:
I expected double dashes in markup (`--`) to render as an em-dash (`—`), but they don't. They just appear as two dashes. Is that intentional? If so that's fine, I just haven't seen that use.

withinfocus (approved):
Looking good to me! Yeah, the double dash is separate from the long dash, but in some renderings this may be automated to be the same.

PR #259, opened 2023-12-15
update client arch docs for team-owned libs

willmartian:
## Objective

Update the web client architecture docs to describe our updated approach to team-owned libs.

### DIFF
--- a/custom-words.txt
+++ b/custom-words.txt
@@ -5,6 +5,7 @@
 Bitwarden
 bytemark
 CODEOWNERS
+codeownership
 CQRS
 dockerized
 Gitter
--- a/docs/architecture/clients/index.md
+++ b/docs/architecture/clients/index.md
@@ -15,13 +15,55 @@ The mono-repository root directory contains three main folders.
 - `bitwarden_license` - Bitwarden Licensed version of the web vault.
 - `libs` - Shared code between the different applications.
 
-`libs` contains the following projects.
+## Libs
 
-- `Common` - Common code shared between all the clients including CLI.
-- `Angular` - Angular specific code used by all the visual clients.
-- `Components` - Angular Components Library.
-- `Node` - Used to be shared code for CLI and Directory Connector CLI, but since directory connector
+Historically, much of the feature logic of Bitwarden has existed within the `apps` directories, only
+being moved to `libs` (usually `libs/common` or `libs/angular`) if it needed to be shared between
+multiple clients.
+
+We are attempting to move to a more modular architecture by creating additional libs that are more
+feature-focused and more accurately convey team-codeownership.
+
+- `common` - Common code shared between all the clients including CLI.
+- `angular` - Low-level Angular specific code used by all the visual clients. Code that is more tied
+  to Bitwarden business logic should be moved to a team-owned lib.
+- `node` - Used to be shared code for CLI and Directory Connector CLI, but since directory connector
   no longer uses the same version of libs this module is deprecated.
+- `admin-console` - Code owned by the Admin Console team.
+- `auth` - Code owned by the Auth team.
+- `billing` - Code owned by the Billing team.
+- `components` - Angular implementation of the Bitwarden design system.
+  [See more.](https://components.bitwarden.com/)
+- `exporter` - Code related to exporting; owned by the Tools team.
+- `importer` - Code related to importing; owned by the Tools team.
+- `platform` - Code owned by the Platform team.
+- `vault` - Code owned by the Vault team.
+
+For more on this approach, check out the
+[Nx documentation](https://nx.dev/concepts/more-concepts/applications-and-libraries):
+
+> place 80% of your logic into the `libs/` folder and 20% into `apps/` We are doing the opposite of
+> this right now. What code should be moved from `apps/` to `libs/`?
+
+- At this time, any code in the `/apps/web/*` that doesn't have a tight coupling with the web client
+  can be moved into a lib. For components, this means they should only be the CL, no global CSS or
+  anything exported from web-specific modules.
+
+> Having a dedicated library project is a much stronger boundary compared to just separating code
+> into folders, though. Each library has a so-called "public API", represented by an index.ts barrel
+> file. This forces developers into an "API thinking" of what should be exposed and thus be made
+> available for others to consume, and what on the others side should remain private within the
+> library itself.
+
+An existing example of this pattern is `@bitwarden/components`:
+
+- It is used by multiple apps and other libs
+- It manages a clear public and private API boundary through its barrel file.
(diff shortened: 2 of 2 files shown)


custom-words.txt
```diff
@@ -5,6 +5,7 @@
 Bitwarden
 bytemark
 CODEOWNERS
+codeownership
```
withinfocus, line 8: ⚠️ We have this as two words everywhere I've seen.

willmartian: 👍 Reverted in [1603713](https://github.com/bitwarden/contributing-docs/pull/259/commits/1603713e09f0bca5c6f0f1d70eab11b577f7d1f9)

docs/architecture/clients/index.md
```diff
@@ -15,13 +15,55 @@ The mono-repository root directory contains three main folders.
...
-- `Components` - Angular Components Library.
-- `Node` - Used to be shared code for CLI and Directory Connector CLI, but since directory connector
+Historically, much of the feature logic of Bitwarden has existed within the `apps` directories, only
+being moved to `libs` (usually `libs/common` or `libs/angular`) if it needed to be shared between
+multiple clients.
+
+We are attempting to move to a more modular architecture by creating additional libs that are more
+feature-focused and more accurately convey team code ownership.
+
+- `common` - Common code shared between all the clients including CLI.
```
withinfocus, line 27: ⛏️ I like double-dashing or using colons for this kind of list item separation. We also try to end bulleted lists without periods.

docs/architecture/clients/index.md
```diff
@@ -15,13 +15,55 @@ The mono-repository root directory contains three main folders.
...
+
+> Having a dedicated library project is a much stronger boundary compared to just separating code
+> into folders, though. Each library has a so-called "public API", represented by an index.ts barrel
+> file. This forces developers into an "API thinking" of what should be exposed and thus be made
+> available for others to consume, and what on the others side should remain private within the
+> library itself.
+
+An existing example of this pattern is `@bitwarden/components`:
+
+- It is used by multiple apps and other libs
```
withinfocus, line 60: ⛏️ Some end in periods, some don't.

docs/architecture/clients/index.md
```diff
@@ -15,13 +15,55 @@ The mono-repository root directory contains three main folders.
...
+- `importer` - Code related to importing; owned by the Tools team.
+- `platform` - Code owned by the Platform team.
+- `vault` - Code owned by the Vault team.
+
+For more on this approach, check out the
+[Nx documentation](https://nx.dev/concepts/more-concepts/applications-and-libraries):
+
+> place 80% of your logic into the `libs/` folder and 20% into `apps/`
+
+We are doing the opposite of this right now. What code should be moved from `apps/` to `libs/`? At
```
withinfocus, line 47: ❓ Do we propose questions to ourselves in this way? I wonder if it's better to be more statement-based alone.

withinfocus (commented):
Some more thoughts. Would still like Gibson or Oscar to take a read.

docs/architecture/clients/index.md
```diff
@@ -15,13 +15,55 @@ The mono-repository root directory contains three main folders.
...
+this time, any code in the `/apps/web/*` that doesn't have a tight coupling with the web client can
+be moved into a lib. For components, this means they should only be the CL, no global CSS or
```
JaredSnider-Bitwarden, line 49: Question: 

> any code in the `/apps/web/*` that doesn't have a tight coupling with the web client can
be moved into a lib

Does this not apply to any client code - not just web?

JaredSnider-Bitwarden, line 49: ⚠️ Unclear text:
> For components, this means **they should only be the CL**, no global CSS or
anything exported from web-specific modules.

Are you saying something like: "For components, they should only be *in* the component library with no global CSS or anything exported from web-specific modules"

willmartian: Yes, thank you. Fixed in [8ce7a11](https://github.com/bitwarden/contributing-docs/pull/259/commits/8ce7a11d3657b49f87303d5efcf2df87e1c6111f).

willmartian: I rewrote this section in [8ce7a11](https://github.com/bitwarden/contributing-docs/pull/259/commits/8ce7a11d3657b49f87303d5efcf2df87e1c6111f), lemme know if this is better:

> We are doing the opposite of this right now. The code that remains in `apps/` should be primarily
> concerned with bootstrapping the application and consuming and configuring code exported from
> `libs/`. Over time, code not meeting this criteria should be moved from `apps/` to `libs/`.
> 
> Any code in the `apps/` that doesn't have a tight coupling with the client can be moved into a lib:
> if an exported member is wholly dependent on other libs but not `apps/`, it can likely be moved to a
> lib itself. (More concretely, this bars components that rely on global client-specific CSS, but not
> components built with Tailwind and the CL.)

docs/architecture/clients/index.md
```diff
@@ -10,44 +10,49 @@ and a single [Git repository](https://github.com/bitwarden/clients).
 
 The mono-repository root directory contains three main folders.
 
-- `apps` - Our different application specific code, consists of `web`, `browser`, `desktop` and
+- `apps`: Our different application specific code, consists of `web`, `browser`, `desktop` and
   `cli`.
-- `bitwarden_license` - Bitwarden Licensed version of the web vault.
-- `libs` - Shared code between the different applications.
+- `bitwarden_license`: Bitwarden Licensed version of the web vault.
+- `libs`: Shared code between the different applications.
```
withinfocus, line 16: ⛏️ Periods are at the end of these.

withinfocus (commented):
Looking good from my perspective.

docs/architecture/clients/index.md
```diff
@@ -10,18 +10,65 @@ and a single [Git repository](https://github.com/bitwarden/clients).
...
-`libs` contains the following projects.
-
-- `Common` - Common code shared between all the clients including CLI.
-- `Angular` - Angular specific code used by all the visual clients.
-- `Components` - Angular Components Library.
-- `Node` - Used to be shared code for CLI and Directory Connector CLI, but since directory connector
-  no longer uses the same version of libs this module is deprecated.
+- `apps`: Our different application specific code, consists of `web`, `browser`, `desktop` and
+  `cli`
+- `bitwarden_license`: Bitwarden Licensed version of the web vault
```
Hinton, line 15: While currently only the web vault it might cover other areas later.

```suggestion
- `bitwarden_license/web-vault`: Bitwarden Licensed version of the web vault
```

withinfocus (commented):
Lint, but this seems good.

PR #269, opened 2024-01-05
First pass at derived state documentation

MGibson1:
## Objective

Adding documentation for derived state to the state provider deep dive.

I moved this to a separate file because it felt optional and different enough to warrant some extra separation.

### DIFF
--- a/docs/architecture/deep-dives/state/derived-state.md
+++ b/docs/architecture/deep-dives/state/derived-state.md
@@ -7,8 +7,8 @@ date when the underlying observable state changes.
 ## `DeriveDefinition`
 
 Derived state has all of the same issues with storage and retrieval that normal state does. Similar
-to `KeyDefinition`, Derived state depends on `DeriveDefinition`s to define magic string keys to
-store and retrieve data from a cache. Unlike normal state, Derived state is always stored in memory.
+to `KeyDefinition`, derived state depends on `DeriveDefinition`s to define magic string keys to
+store and retrieve data from a cache. Unlike normal state, derived state is always stored in memory.
 It still takes a `StateDefinition`, but this is used only to define a namespace for the derived
 state, the storage location is ignored. _This can lead to collisions if you use the same key for two
 different derived state definitions in the same namespace._
@@ -49,7 +49,7 @@ often built from existing state and allows you to create a definition from an ex
   `DeriveDefinition`s require have a `deserializer` function that is used to convert the stored data
   back into the `TTo` type.
 - `derive` - A function that takes the current state and returns the derived state. This function
-  takes two parameters
+  takes two parameters:
   - `from` - The latest value of the parent state.
   - `deps` - dependencies used to instantiate the derived state. These are provided when the
     `DerivedState` class is instantiated. This object should contain all of the application runtime
--- a/docs/architecture/deep-dives/state/derived-state.md
+++ b/docs/architecture/deep-dives/state/derived-state.md
@@ -1,0 +1,129 @@
+# Derived State
+
+It is common to need to cache the result of expensive work that does not represent true alterations
+in application state. Derived state exists to store this kind of data in memory and keep it up to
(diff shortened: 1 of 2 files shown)


docs/architecture/deep-dives/state/derived-state.md
```diff
@@ -0,0 +1,129 @@
+# Derived State
+
+It is common to need to cache the result of expensive work that does not represent true alterations
+in application state. Derived state exists to store this kind of data in memory and keep it up to
+date when the underlying observable state changes.
+
+## `DeriveDefinition`
+
+Derived state has all of the same issues with storage and retrieval that normal state does. Similar
+to `KeyDefinition`, Derived state depends on `DeriveDefinition`s to define magic string keys to
```
withinfocus, line 10:
```suggestion
to `KeyDefinition`, derived state depends on `DeriveDefinition`s to define magic string keys to
```

docs/architecture/deep-dives/state/derived-state.md
```diff
@@ -0,0 +1,129 @@
...
+store and retrieve data from a cache. Unlike normal state, Derived state is always stored in memory.
```
withinfocus, line 11:
```suggestion
store and retrieve data from a cache. Unlike normal state, derived state is always stored in memory.
```

docs/architecture/deep-dives/state/derived-state.md
```diff
@@ -0,0 +1,129 @@
+# Derived State
+
+It is common to need to cache the result of expensive work that does not represent true alterations
+in application state. Derived state exists to store this kind of data in memory and keep it up to
+date when the underlying observable state changes.
+
+## `DeriveDefinition`
```
withinfocus, line 7: ⛏️ and it may be too late here, but why not "derived" for the object name? This would coincide with the "derived" state lower down. Or is this a "derivation" definition?

docs/architecture/deep-dives/state/derived-state.md
```diff
@@ -0,0 +1,129 @@
...
+### `DeriveDefinitionOptions`
+
+[The `DeriveDefinition` section](#deriveDefinitionFactories) specifies a third parameter as
+`_DeriveOptions_`, which is used to fully specify the way to transform `TFrom` to `TTo`.
+
+- `deserializer` - For the same reasons as [Key Definition Options](#keydefinitionoptions),
+  `DeriveDefinition`s require have a `deserializer` function that is used to convert the stored data
+  back into the `TTo` type.
+- `derive` - A function that takes the current state and returns the derived state. This function
+  takes two parameters
```
withinfocus, line 52:
```suggestion
  takes two parameters:
```

docs/architecture/deep-dives/state/derived-state.md
```diff
@@ -0,0 +1,129 @@
...
+  - `from` - The latest value of the parent state.
+  - `deps` - dependencies used to instantiate the derived state. These are provided when the
+    `DerivedState` class is instantiated. This object should contain all of the application runtime
+    dependencies for transform the from parent state to the derived state.
+- `cleanupDelayMs` (optional) - Takes the number of milliseconds to wait before cleaning up the
```
withinfocus, line 57: ❓ What's the underlying purpose for this?

docs/architecture/deep-dives/state/derived-state.md
MGibson1: The thinking was just that the state has already been done, hence past tense while the definition is describing how to derive.

docs/architecture/deep-dives/state/derived-state.md
MGibson1: This is effectively the lifetime of derived values to live in memory once observers unsubscribe.

If you have some data that is expensive to derive, you may want to increase this to minimize the chance that data must be re-calculated.

withinfocus (approved):
Makes sense!

PR #274, opened 2024-01-09
[PM-5649] Remove clsx dependency and unused components

jprusik:
## Objective

In looking into updating the `clsx` dependency, I noticed it was only used for a single component (`HomepageFeatures`), which was part of Docusaurus initialization and not in use in the contributing-docs site. Removing the component allows use to remove the `clsx` dependency

Also, resolve a Docusaurus warning about a broken anchor


Hinton:
Looks safe to remove. It was probably part of the initial scaffolding.

PR #281, opened 2024-01-21
Add instructions for feature flag removal

trmartin4:
## Objective

Added instructions on feature flag removal.

### DIFF
--- a/docs/contributing/feature-flags.md
+++ b/docs/contributing/feature-flags.md
@@ -276,18 +276,24 @@ our API, it is important that each feature flag be removed in the appropriate se
 First, remove all business logic that relies on the flag from both client and server code. This
 includes _all_ references in the client codebase, and also any business logic on the server that
 checks the flag value. This does _not_ include removing the flag from the `FeatureFlagKeys` on the
-server. We must leave this here so that old clients who have not updated continue to be served a
-`true` value when querying for the flag.
+server -- we must leave this here so that old clients who have not updated continue to be served the
+correct "on" value when querying for the flag.
 
-This code should then be deployed to all clients and to the server. Once this has been deployed,
-clients have updated, and the self-hosted customer has updated their deployment to the latest
-version, self-hosted customers will be able to access the feature (see
-[Self-hosted considerations](#self-hosted-considerations) below).
+This code should then be deployed to all clients and to the server.
 
-After waiting the requisite number of client releases to ensure backward compatibility, we can then
-take the next step and completely remove the feature flag from the server codebase. This can be done
-by removing the flag value from the `FeatureFlagKeys`. This should then be deployed to the server to
-complete the removal process.
+:::tip Self-hosted customers
+
+Once the code referencing the flag has been removed and clients have updated, self-hosted customers
+can update to the latest version to begin using the feature. See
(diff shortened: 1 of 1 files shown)


docs/contributing/feature-flags.md
```diff
@@ -262,11 +262,34 @@ to Jira helps create a history of the feature and there are copious logs and aud
...
+
+### Unwinding a feature flag
+
+Due to the complexity of the different client deployments and how we expose feature flags through
+our API, it is important that each feature flag be removed in the appropriate sequence.
+
+First, remove all business logic that relies on the flag from both client and server code. This
+includes _all_ references in the client codebase, and also any business logic on the server that
+checks the flag value. This does _not_ include removing the flag from the `FeatureFlagKeys` on the
+server. We must leave this here so that old clients who have not updated continue to be served a
```
withinfocus, line 279:
```suggestion
server -- we must leave this here so that old clients who have not updated continue to be served a
```

docs/contributing/feature-flags.md
```diff
@@ -262,11 +262,34 @@ to Jira helps create a history of the feature and there are copious logs and aud
...
+`true` value when querying for the flag.
```
withinfocus, line 280: 🎨 Would be good to open up the language here a bit. Not all flags are Booleans, so something like "on" is better than a code-like `true`.

docs/contributing/feature-flags.md
```diff
@@ -262,11 +262,34 @@ to Jira helps create a history of the feature and there are copious logs and aud
...
+
+This code should then be deployed to all clients and to the server. Once this has been deployed,
+clients have updated, and the self-hosted customer has updated their deployment to the latest
```
withinfocus, line 283: 💭 You clarify it more below, but "clients have updated" is an open-ended term that has a long tail to it. I wonder if we should rephrase a bit to this being about how we have seen a significant rollout to clients.

🎨 I don't quite understand the language for the part about self-host here. I'd instead have a one-liner as an independent paragraph stating how self-host only gets features once the flag checks are removed. Could even be emphasized in a block.

withinfocus (changes requested):

docs/contributing/feature-flags.md
trmartin4: Updated in [`1f3ee4a` (#281)](https://github.com/bitwarden/contributing-docs/pull/281/commits/1f3ee4a29803f6c3cd7363e79a250653ec5686ee).

docs/contributing/feature-flags.md
trmartin4: [`1f3ee4a` (#281)](https://github.com/bitwarden/contributing-docs/pull/281/commits/1f3ee4a29803f6c3cd7363e79a250653ec5686ee)

docs/contributing/feature-flags.md
trmartin4: I've updated in [`1f3ee4a` (#281)](https://github.com/bitwarden/contributing-docs/pull/281/commits/1f3ee4a29803f6c3cd7363e79a250653ec5686ee).  This change assumes that our backward compatibility documentation has been updated.

withinfocus (approved):
Love it!

PR #289, opened 2024-01-31
Clarify instructions for local SAML SSO

trmartin4:
## Objective

With https://github.com/bitwarden/clients/pull/7117, we added an SSO configuration option (defaulting to enabled) that  creates a specific Entity ID in the URL.

In the setup documentation, the URL didn't reflect that change.  Despite the instructions to copy from the values in the web client, the presence of the "copy" UI made it very easy to copy the incorrect URL and set the environment variable incorrectly.

Instead of adding "{yourOrgIdHere}" as was already in the ACS URL, I opted to completely remove any reference to the URLs themselves in the script, so that users will use the values in the UI.  This will prevent us from having to keep the documentation in sync with any future URL structure changes.

### DIFF
--- a/docs/getting-started/server/sso/local.md
+++ b/docs/getting-started/server/sso/local.md
@@ -33,7 +33,8 @@ This uses
     ```
 
 6.  Open your `.env` file and set the following environment variables using the "SP Entity ID" and
-    "Assertion Consumer Service (ACS) URL" values from the SSO configuration page opened in step #2:
+    "Assertion Consumer Service (ACS) URL" values from the SSO configuration page opened in step #4
+    above:
 
     ```bash
     IDP_SP_ENTITY_ID={SP Entity ID}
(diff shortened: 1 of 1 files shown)


docs/getting-started/server/sso/local.md
```diff
@@ -13,31 +13,31 @@ This uses
...
 5.  Open a new terminal and navigate to the `dev` folder in your server repository, e.g.
 
     ```bash
     cd ~/Projects/server/dev
     ```
 
-6.  Open your `.env` file and set the following environment variables, based on the "SP Entity ID"
-    and "Assertion Consumer Service (ACS) URL" values on the web vault SSO configuration page:
+6.  Open your `.env` file and set the following environment variables using the "SP Entity ID" and
+    "Assertion Consumer Service (ACS) URL" values from the SSO configuration page opened in step #2:
```
JaredSnider-Bitwarden, line 36: Question:  What step are you referencing in: `opened in step #2` ?  Step 2 is `Tick the "Allow SSO authentication" box.` within this section?

trmartin4: Oops, I put #2 instead of #4.  I'e fixed it.

JaredSnider-Bitwarden (approved):
Thank you for updating this! LGTM!

djsmith85 (approved):

PR #296, opened 2024-02-05
Added optional constructor parameter and updated mapping instructions.

trmartin4:
## Objective

Added additional documentation for new constructor parameter as well as a mapping table to help with migration.

📓 Note the destination of this PR is @justindbaur's larger body of work updating the state provider framework.  I wanted to make sure I had a review before it made it into this larger set of changes, hence the separate PR.


justindbaur (approved):
Thank you so much for adding this.

PR #310, opened 2024-02-23
Rework ADR 12

Hinton:
## Objective

Rework the abstraction suffix to something less verbose. My aspiration is to settle this discussion and do a quick follow up with a PR to update the codebase.


MGibson1 (dismissed):
Hopefully this one takes...

docs/architecture/adr/0012-angular-filename-convention.md
```diff
@@ -45,16 +45,27 @@ At Bitwarden we also use a couple of more types:
...
+the Angular Dependency Injection. In order to make code useable in both Angular and non Angular
+based clients we generally use abstract classes to define the interface. Ideally we would use
+_interfaces_ but TypeScript interfaces cannot be used to wire up dependency injection in JavaScript.
+
+All consumers will use the abstract class as a parameter in their constructor which will be manually
+wired up in the CLI and use Angular dependency injection in the Angular clients. In the case a
+dependency is only used in the Angular client, the abstract class can be omitted and the
+implementation referenced directly using `@Injectable`.
+
+To avoid naming conflicts, it's forbidden to name classes with the same name as the abstract class.
```
audrey-jensen, line 64: 💭 It isn't clear to me why a naming conflict is a problem. Why not instead use `import as` to avoid the conflict?

Note: this is not saying I disagree, only that this ADR does not assert why we should opt to use one over the other. Indeed, it seems out of place to talk about class naming conventions in an ADR concerning file naming conventions.

docs/architecture/adr/0012-angular-filename-convention.md
```diff
@@ -45,16 +45,27 @@ At Bitwarden we also use a couple of more types:
...
+Depending on the class usage the following prefixes are allowed:
+
+- `Default`: Used for the default implementation of an abstract class.
+- `Web`, `Browser`, `Desktop` and `Cli` for platform specific implementations.
```
audrey-jensen, line 68: 🤔 Has repetition in the file path been considered? I'd think it's safe, and perhaps even desirable for search purposes, to omit the platform-specific prefix from a file (or most differentiators) when that information is already captured in the module name. It's like a DRY microcosm.

A toy example based on some of the things I've seen:

WebPasswordGeneratorComponent
"web/tools/generator/password.component.html"
vs
"web/tools/generator/web-password-generator.component.html"

docs/architecture/adr/0012-angular-filename-convention.md
Hinton: There's nothing inherently wrong with name conflicts, it's more an area of confusion and additional overhead to consider when resolving automatic imports in editors, and performing code reviews.

The original reason this was part of the ADR was due to the initial iteration using a `.implementation` suffix, which later became `.abstraction`. The implicit effect of this rule is that the file will be named `default-`.

It's true this isn't directly related to this ADR and we could remove it and just keep it in the presentation document.

docs/architecture/adr/0012-angular-filename-convention.md
Hinton: Not really. The occurrences where we require different services in clients tends to be fairly small and should really only be applicable to services.

The example seems to be an old component that uses inheritance? Following more modern composition it wouldn't require a client specific prefix.

docs/architecture/adr/0012-angular-filename-convention.md
audrey-jensen: I think it's important to include that justification in the ADR. As is, the PR reads as "Do it because I said so" when it's really "Do it because `import as` is considered harmful."

docs/architecture/adr/0012-angular-filename-convention.md
audrey-jensen: This ADR doesn't seem to limit itself to services and talks about domain objects and so forth, so I'm interpreting it as applying across all of the code.

As far as the convention goes, there's a tendency towards abstraction in tools code. Components increasingly have base classes that provide reactive bindings, with client-specific code containing little more than a `@Component()` that bridges the bindings to a client-specific template. Since the component library work is presently limited to the web environment, the future state of this convention is unclear. 

We want a single responsive component, but the patterns we'd use to create them are still nascent, when they're realized at all. What can be said is that this is often the latest code available. They're not old components; they're complex, self-contained smart components.

docs/architecture/adr/0012-angular-filename-convention.md
```diff
@@ -45,16 +45,27 @@ At Bitwarden we also use a couple of more types:
...
+## Abstract & Default Implementations
+
+The Bitwarden clients codebase serves multiple clients, one of which is node based and doesn't use
+the Angular Dependency Injection. In order to make code useable in both Angular and non Angular
+based clients we generally use abstract classes to define the interface. Ideally we would use
+_interfaces_ but TypeScript interfaces cannot be used to wire up dependency injection in JavaScript.
+
+All consumers will use the abstract class as a parameter in their constructor which will be manually
+wired up in the CLI and use Angular dependency injection in the Angular clients. In the case a
+dependency is only used in the Angular client, the abstract class can be omitted and the
```
willmartian, line 61:
```suggestion
dependency is only used in Angular contexts, the abstract class can be omitted and the
```

We have multiple clients that use Angular, but only only client is *entirely Angular*

willmartian (dismissed):
Good change. The piece that needs to be least verbose is the abstract class name, because it is referenced most often as a dependency injection token. The consumer should be agnostic as to whether they are injecting an abstract class or some implementation. 

I'm not worried about specific class implementations being more verbose because they will only need to be referenced in a `providers` array.

Hinton:
I've reworked this to primarily affect the angular architecture documentation, and remove the Abstraction section completely from the ADR. I believe this captures some of the review feedback.

audrey-jensen (approved):

MGibson1 (approved):

PR #311, opened 2024-02-26
Consistently use capitalization of 'macOS' in documentation

KatherineInCode:
## Objective

Throughout the documentation, we were sometimes referring to the Mac operating system as "MacOS" and "macOS". The recommendation in the [Apple Style Guide](https://support.apple.com/en-my/guide/applestyleguide/apsg72b28652/1.0/web/1.0#apd246e83209) is "macOS", and while that document is intended primarily for Apple communications, it seems reasonable for us to match their branding.

This change updates this accordingly.

### DIFF
--- a/docs/getting-started/clients/browser/biometric.mdx
+++ b/docs/getting-started/clients/browser/biometric.mdx
@@ -25,7 +25,7 @@ _Supported:_
 _Not Supported:_
 
 - Firefox Versions 86 and earlier
-- Side-loaded MacOS Desktop Apps
+- Side-loaded macOS Desktop Apps
 - Linux OS
 
 ## General Setup Steps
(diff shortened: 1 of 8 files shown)


docs/getting-started/clients/browser/biometric.mdx
```diff
@@ -25,7 +25,7 @@ _Supported:_
 _Not Supported:_
 
 - Firefox Versions 86 and earlier
-- Side-loaded MacOS Desktop Apps
+- Side-loaded macOS Desktop Apps
```
withinfocus, line 28: ⛏️ "Desktop Apps" shouldn't be capitalized either, or "Versions" ... probably some more of these.

withinfocus (approved):
Nice quality change! Left one ⛏️ but it's technically out of scope for this change.

docs/getting-started/clients/browser/biometric.mdx
KatherineInCode: Good call. I'll make notes to circle back to that, possibly in a broader "make more capitalization more consistent" change—I think there were a few other things I noticed along those lines when I was reading through recently

PR #312, opened 2024-02-27
Add note to EF page for reloading secrets

KatherineInCode:
## Objective

When I was trying to get my local Bitwarden server working with an EF database, I ran into the problem that changing the `secrets.json` file alone was insufficient. After a bit of time, I figured out that calling the `setup_secrets` script again fixed the issue.

It therefore seems reasonable to me to include a note here to remind people to run that script if modifying properties to run the server with an EF.

### Caveats
- It is possible that this is not actually necessary, and there is some other better solution, in which case the docs should be updated to that effect
- I only used `pwsh setup_secrets.ps1 -clear`, and that worked; it's possible the `-clear` flag is unnecessary, though it seems reasonable especially if the `databaseProvider` property might be removed (to return to MSSQL)

### DIFF
--- a/docs/getting-started/clients/browser/biometric.mdx
+++ b/docs/getting-started/clients/browser/biometric.mdx
@@ -25,7 +25,7 @@ _Supported:_
 _Not Supported:_
 
 - Firefox Versions 86 and earlier
-- Side-loaded MacOS Desktop Apps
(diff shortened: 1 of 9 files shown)


KatherineInCode:
I screwed up my branches. Sorry. Will fix.

PR #313, opened 2024-02-27
Clarify that changing secrets.json requires reloading the secrets

KatherineInCode:
## Objective

When I was trying to get my local Bitwarden server working with an EF database, I ran into the problem that changing the `secrets.json` file alone was insufficient. After a bit of time, I figured out that calling the `setup_secrets` script again fixed the issue.

It therefore seems reasonable to me to include a note here to remind people to run that script if modifying properties to run the server with an EF.

### Caveats
- It is possible that this is not actually necessary, and there is some other better solution, in which case the docs should be updated to that effect
- I only used `pwsh setup_secrets.ps1 -clear`, and that worked; it's possible the `-clear` flag is unnecessary, though it seems reasonable especially if the `databaseProvider` property might be removed (to return to MSSQL)

### DIFF
--- a/docs/getting-started/server/database/ef/index.mdx
+++ b/docs/getting-started/server/database/ef/index.mdx
@@ -135,7 +135,7 @@ Source path. Use the file location selected in [Database Setup](#database-setup)
 :::note
 
 After making changes to your `secrets.json` file, remember to run `pwsh setup_secrets.ps1 -clear` to
-make sure the changes are incorporated.
+make sure the changes take effect.
 
 :::
 
--- a/docs/getting-started/server/database/ef/index.mdx
+++ b/docs/getting-started/server/database/ef/index.mdx
@@ -132,6 +132,13 @@ Source path. Use the file location selected in [Database Setup](#database-setup)
 </TabItem>
 </Tabs>
 
+:::note
+
+After making changes to your `secrets.json` file, remember to run `pwsh setup_secrets.ps1 -clear` to
+make sure the changes are incorporated.
+
+:::
+
 ### Migrations
 
 <Tabs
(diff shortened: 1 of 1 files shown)


docs/getting-started/server/database/ef/index.mdx
```diff
@@ -132,6 +132,13 @@ Source path. Use the file location selected in [Database Setup](#database-setup)
 </TabItem>
 </Tabs>
 
+:::note
+
+After making changes to your `secrets.json` file, remember to run `pwsh setup_secrets.ps1 -clear` to
+make sure the changes are incorporated.
```
eliykat, line 138: Good tip! A slight nit / suggestion on wording:

```suggestion
After making changes to your `secrets.json` file, remember to run `pwsh setup_secrets.ps1 -clear` to
make sure the changes take effect.
```

KatherineInCode: Good call!

docs/getting-started/server/database/ef/index.mdx
```diff
@@ -132,6 +132,13 @@ Source path. Use the file location selected in [Database Setup](#database-setup)
 </TabItem>
 </Tabs>
 
+:::note
+
+After making changes to your `secrets.json` file, remember to run `pwsh setup_secrets.ps1 -clear` to
```
withinfocus, line 137:
```suggestion
After making changes to your `secrets.json` file, remember to run `pwsh setup_secrets.ps1 -clear` so
```

docs/getting-started/server/database/ef/index.mdx
```diff
@@ -132,6 +132,13 @@ Source path. Use the file location selected in [Database Setup](#database-setup)
...
+make sure the changes take effect.
```
withinfocus, line 138:
```suggestion
that the changes take effect.
```

withinfocus (changes requested):
⛏️ Repetitive "to" so one minor suggestion.

withinfocus (approved):
Looking good, thanks!

PR #316, opened 2024-03-08
Add the SDK to clients

Hinton:

### DIFF
--- a/docs/getting-started/mobile/android/index.md
+++ b/docs/getting-started/mobile/android/index.md
@@ -2,7 +2,7 @@
 
 ## Requirements
 
-Before you start, you should have the recommended [Tools and Libraries](../../../tools/index.md)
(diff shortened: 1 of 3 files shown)


docs/getting-started/clients/sdk/index.md
differsthecat, line 1: Does this make sense under clients? It's not really a client in my eyes, I would think it should be its own parent page.

Hinton: Maybe, it's limited to the clients though. Maybe we're better of breaking out mobile and sdk from clients and let it be focused on the clients repository?

differsthecat: Ah yeah, mobile is in there too; I do think it would make sense to break them up more by repo, especially since the [main landing page](https://contributing.bitwarden.com/getting-started/clients/) only mentions the client's repo

differsthecat (approved):
Looks good, thank you!

PR #317, opened 2024-03-08
New ADR statuses

withinfocus:
## Objective

Uses new statuses for ADRs. After some review of what others do with them and how we have a number ourselves that we consider ratified but not necessarily appropriate to declare "done", it's better to represent acceptance and future state.

### DIFF
--- a/docs/architecture/adr/0004-refactor-state-service.md
+++ b/docs/architecture/adr/0004-refactor-state-service.md
@@ -1,6 +1,6 @@
 ---
 adr: "0004"
-status: In progress
+status: Accepted
 date: 2022-06-30
 tags: [clients, angular]
 ---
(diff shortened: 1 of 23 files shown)


docs/architecture/adr/index.mdx
```diff
@@ -23,9 +23,16 @@ The AD will also serve as a base for proposing and planning technical debt.
```
Hinton, line 23: Could we add some text to the above section explaining that an ADR is expected to update the architecture documentation to reflect the latest status?

docs/architecture/adr/0004-refactor-state-service.md
Hinton, line 1: @MGibson1 the state and mv3 adrs are probably in need to be supserseeded with our current strategy.

withinfocus: Will a new ADR be released then? Otherwise they could just be edits to what's accepted.

PR #320, opened 2024-03-14
Updated self-host git command

trmartin4:
## Objective

Fixed self-host git command to use `https` URL, matching the cloud instructions.


withinfocus (commented):
Hmm, I think we should go the other way and promote SSH usage everywhere.

eliykat:
This text has been deleted in #328 so this PR can be closed.

PR #326, opened 2024-03-20
Describe advanced local feature flag configuration

MGibson1:
## Objective

I had cause to add test flag differences between authenticated and unauthenticated contexts.

In a local system, this means needing to expand our basic JSON definition considerably to define segments, rules, and flags. This documents the basics of how that works and provides an example to work from should others need to do the same.

### DIFF
--- a/docs/contributing/feature-flags.md
+++ b/docs/contributing/feature-flags.md
@@ -1,8 +1,3 @@
----
-sidebar_custom_props:
-  access: bitwarden
----
-
 # Feature Flags
 
 ## Background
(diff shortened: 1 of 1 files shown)


docs/contributing/feature-flags.md
```diff
@@ -1,8 +1,3 @@
----
```
withinfocus, line 1: ❓ Why open this up to everyone? We do not want to encourage usage of this except internally.

docs/contributing/feature-flags.md
```diff
@@ -97,6 +92,107 @@ configuration setting can be used to override it. The file must be present befor
 solution, but once there you can change the file contents and see immediate results in running /
 debugging code.
 
+<details>
+  <summary>Context-aware feature flag JSON</summary>
+  <div>
+    <div>
+      The `flags.json` file can also define flags which respond to user context. Currently, only `UserId`and the `OrganizationId` of all organizations a user belongs to are included in our feature flagging context. The syntax for defining context-aware flags amounts to defining a `flag` object which specifies `variations` values and `rules` which are evaluated against the user context. A `fallthrough` object is also available to specify a default variation. Rules are represented by `segments`, which are defined in the same file.
```
withinfocus, line 99:
```suggestion
      The `flags.json` file can also define flags that respond to user context. Currently, only `UserId` and the `OrganizationId` of all organizations a user belongs to are included in our feature flagging context. The syntax for defining context-aware flags amounts to defining a `flag` object that specifies `variations` values and `rules` which are then evaluated against the user context. A `fallthrough` object is also available to specify a default variation. Rules are represented by `segments`, which are defined in the same file.
```

withinfocus: ⛏️ This is also violating our line lengths since it's in custom HTML.

docs/contributing/feature-flags.md
```diff
@@ -97,6 +92,107 @@ configuration setting can be used to override it. The file must be present befor
...
+
+      A Feature flag may not appear in both the `flags` object and in the `flagValues` object.
```
withinfocus, line 101:
```suggestion
      A feature flag may not appear in both the `flags` object and in the `flagValues` object.
```

docs/contributing/feature-flags.md
```diff
@@ -97,6 +92,107 @@ configuration setting can be used to override it. The file must be present befor
...
+
+      Note that UserId is referred to as `key` in the `clauses` object while defining a segment.
```
withinfocus, line 103:
```suggestion
      Note that `UserId` is referred to as `key` in the `clauses` object while defining a segment.
```

docs/contributing/feature-flags.md
withinfocus: ❌ We support an addition service account context as well that's omitted.

withinfocus (changes requested):
Pretty advanced use case to show, but perhaps it will help advanced needs.

docs/contributing/feature-flags.md
MGibson1: Good point, we actually have three `contextKinds` now that I look closer. I'll expand to include descriptions of data on each

docs/contributing/feature-flags.md
MGibson1: The metric I use for internal vs external is whether or not special access is required. The feature flags is in the AGPL licensed section of the codebase and not blocked by any employee-specific secrets.

I don't see why we wouldn't want external contributors feature flagging their contributions, if it makes sense and they are able.

withinfocus: They're not able, since we'd have to register flags in LD. If you think an engineer would connect with them to do that then alright, but it's a bit of a jump from today.

There's sometimes an eagerness to adopt features in the community and what's happened is users start trying to put these developer overrides in and have problems, therefore increasing support needs. I want the impetus to be on our internal engineering team to quickly remove these in a follow-up release instead.

PR #333, opened 2024-04-07
Dependency injection writeup

eliykat:
## Objective

My initial goal here was to document [safeProvider](https://github.com/bitwarden/clients/pull/8206), however I decided that we needed a writeup about our DI patterns in general to properly contextualize it. This was tricky to structure, partly because our patterns in this area are varied and diffuse, so feedback is particularly welcome.


docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,157 @@
...
+automatically configure its dependencies. If a service is used in mixed contexts (e.g.
+`libs/common`), it must not use Angular decorators and its dependencies must be manually configured.
+
+## Configuring DI
+
+### Angular contexts
+
+Angular contexts use ngModules
+[dependency providers](https://angular.io/guide/dependency-injection-providers) to configure DI. For
+example:
```
coroiu, line 46: I think there are some additional subtleties that we should probably mention here. This method of configuring DI services is mainly used for multi-client services. For single-client services and/or angular-only services, we should also mention:

### provideIn
Services that are intended to be used globally can use the newer `provideIn` property which avoids having to create and manage service-modules which can be tricky to manage when exported through feature-modules, usually ending up in multiple instances of the service being created.
```
@Injectable({
    provideIn: "root"
})
```

### Standalone components
Services that do are not stateful and/or are only supposed to be used locally to extract complex logic from a component can be injected directly into standalone components, completely bypassing modules.
```
@Component({
    providers: [SomeService],
    standalone: true
})
```

eliykat: `providedIn` assumes that there is no abstract interface for the service, right? Do we have any clear guidelines/rules as to when that's appropriate? The current practice seems to be "small, Angular-only services don't need abstractions if you don't feel like it" but I'm not really sure why.

Perhaps abstract interfaces are only used to manage different implementations across apps, if a service is for a single app only, it doesn't need one? (This seems to be the approach SM Team have taken, I see `providedIn: "root"` for all their services.)

coroiu: > providedIn assumes that there is no abstract interface for the service, right?

Exactly!

> Perhaps abstract interfaces are only used to manage different implementations across apps, and if a service is for a single app only, it doesn't need one?

This is the same approach I've been using! The component library might also export services without abstractions, not sure if they use `providedIn` though, but they probably should!

eliykat: After discussing this on Slack, I've incorporated your suggestions. Unfortunately there are now a lot of options for how to register a service in Angular - but that reflects our current practice. Open to any further feedback re: structure or being more opinionated in the text.

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+```ts
+// Injected in other modules
+abstract class MyService {
+  externalMethod: () => void;
+}
+
+// Injected in the MyService feature module
+abstract class InternalMyService extends MyService {
+  internalMethod: () => void;
+}
```
coroiu, line 60: 🎨 suggestion: I think we're switching to true function definitions

```suggestion
// Injected in other modules
abstract class MyService {
  abstraction externalMethod(): void;
}

// Injected in the MyService feature module
abstract class InternalMyService extends MyService {
  abstract internalMethod(): void;
}
```

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+  their dependencies
+
+## Services
+
+Services should be organized as follows depending on where they're used:
+
+- `libs/common` for services used by **multiple clients** in **non-Angular contexts (or a mix of
+  Angular and non-Angular contexts)**
+- `libs/angular` for services used by **multiple clients** in **Angular contexts only**
+- `apps/<client-name>` for services used by a single client only
```
coroiu, line 34: 🎨 suggestion: mention the team-specific libraries e.g. `libs/auth`, `libs/vault`

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+service.
+
+## Configuring DI
+
+### Angular modules
+
+Angular contexts generally use ngModules
+[dependency providers](https://angular.io/guide/dependency-injection-providers) to configure DI. For
+example:
+
```
coroiu, line 78: :art: suggestion(non-blocking): Add admonitions and/or consider changing the order of the examples so people don't stop reading after the first code example and think it is the recommended one

```suggestion
Angular contexts generally use ngModules
[dependency providers](https://angular.io/guide/dependency-injection-providers) to configure DI. For
example:

:::danger[BAD]

Do not do this, it is not type safe.

:::

File: docs/architecture/clients/dependency-injection.md [hunk 5]
```diff
@@ -0,0 +1,219 @@
...
+})
+export class MyServicesModule {}
+```
+
+:::note
+
+The `useAngularDecorators` option is not type safe - it is your responsibility to ensure that the
+class actually uses the `@Injectable` decorator.
+
+:::
```
coroiu, line 129: 💭 thought: we should keep an eye on https://github.com/microsoft/TypeScript/issues/4881 which might be able to fix this

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+Our primary service modules require the use of `safeProvider`, otherwise you will receive type
+errors. However, you should use `safeProvider` wherever you are configuring an Angular provider to
+benefit from its type checking.
+
+### Angular root module
+
+Services that will be used globally and which don't have an abstract interface can be registered
+using the `provideIn` property. This avoids having to create and manage service modules. When
+exporting services through feature modules, it can be tricky to manage their lifetime, and it's easy
+to end up with instances of the service being created.
```
coroiu, line 140: typo: 
```suggestion
to end up with multiple instances of the service being created.
```

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+```ts
+@Injectable({
+  provideIn: "root",
+})
+export class MyService {}
+```
+
+### Angular components
+
+Services that do are not stateful and/or are only supposed to be used locally to extract complex
```
coroiu, line 151: typo?
```suggestion
Services that are not stateful and/or are only supposed to be used locally to extract complex
```

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+logic from a component can be injected directly into standalone components, completely bypassing
+modules.
+
+```ts
+@Component({
+  providers: [
+    // This example assumes that MyService uses @Injectable decorators and does not have an abstract interface
+    safeProvider(MyService),
```
coroiu, line 159: ❓ question: interesting use of the `safeProvider` function! Does it do something here or are we just recommending it for consistency?

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+  ],
+  standalone: true,
+})
+export class MyComponent {}
+```
+
+### Non-Angular contexts
+
+Non-Angular contexts do not use a DI framework and must manually instantiate their dependencies **in
```
coroiu, line 168: typo?: 
```suggestion
Non-Angular contexts do not use a DI framework must manually instantiate their dependencies **in
```

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+the correct order**. Instantiating dependencies out of order may result in null values being
+injected into services, causing runtime errors.
```
coroiu, line 170: 💭 thought: This should be fixed by TS strict mode
image

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+
+## DI by client
+
+### Shared Angular DI
+
+`libs/angular` contains `jslib-services.module.ts`, a services module which registers common
```
coroiu, line 176: nit: this section refers to this module as `jslib-services.module.ts` while the ones below seem to use its class name
```suggestion
`libs/angular` contains `JslibServicesModule`, a services module which registers common
```

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+A client may override this DI if a client-specific configuration is required.
+
+### Web
+
+`core.module.ts` contains the main services module for the web client. It imports
+`JslibServicesModule`.
+
+However, the web module heavily uses [feature modules](https://angular.io/guide/feature-modules) to
+divide its code by feature. Feature modules should configure their own DI wherever possible for
+feature-specific (non-global) services.
```
coroiu, line 188: suggestion(non-blocking): preferably in standalone components or using `provideIn: 'root'` and not in the actual feature module

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,219 @@
...
+Desktop DI is split across the following locations:
+
+- `services.module.ts` is the main services module for the Electron renderer process. It imports
+  `JslibServicesModule`
+- `main.ts` manually instantiates services used in the main process. It does not use any dependency
+  injection framework
+
+The main process is primarily used for Electron behavior and native integrations (e.g. storage,
+biometrics, window behavior), so it registers a smaller subset of services. Many services are only
+registered in the renderer process.
```
coroiu, line 215: suggestion: maybe clarify that the `renderer process` is basically the "web page" part of Electron (that is how it is, right..? 😅 )

coroiu (commented):
I gave all of the parts a detailed read-through this time, and while I did have quite a few comments, it's all mostly just polish and I would've ok with merging as-is.

docs/architecture/clients/dependency-injection.md
eliykat: Done!

docs/architecture/clients/dependency-injection.md
eliykat: This is a good call - it made me wonder whether I shouldn't lead with an example that is wrong, but I think it's necessary to build context. The admonitions help clarify which is the good & bad example though.

docs/architecture/clients/dependency-injection.md
eliykat: Nice spot!

docs/architecture/clients/dependency-injection.md
eliykat: yes indeed

docs/architecture/clients/dependency-injection.md
eliykat: Just for consistency, so that we can say "you must always use safeProvider" (and type check this via the `SafeProvider` return type) without having to give exceptions. I agree it doesn't add much otherwise.

Really this is the equivalent of `safeProvider({ provide: MyService, useClass: MyService, useAngularDecorators: true})` but why require all that 😁

docs/architecture/clients/dependency-injection.md
eliykat: Nope! It's saying 2 things:
1. Non-Angular contexts do not use a DI framework
2. Non-Angular contexts must manually instantiate their dependencies

To put it another way, the "and" is a logical "and therefore".

docs/architecture/clients/dependency-injection.md
eliykat: TIL just how strict strict mode is. One day...

docs/architecture/clients/dependency-injection.md
eliykat: It sure is! done

docs/architecture/clients/dependency-injection.md
eliykat: Why not? If the component is standalone, then by definition it doesn't have a module, so advice about a feature module doesn't apply. And if you have a service that is only used in a feature module, I assume you'd want to register it in the more specific location rather than in root. Does this go to your earlier concern about accidentally instantiating duplicate services?

eliykat:
Thanks for all the feedback @coroiu , I've addressed or responded to everything.

Can I please have a review from @bitwarden/dept-architecture as codeowners as well, thanks.

docs/architecture/clients/dependency-injection.md
coroiu: Yeah I agree that it makes sense to have the bad example first, otherwise you won't know what to look for in the good example. So it's mostly just clarifying that this is one of those "Bad / Good" sections. Here are some  examples I looked at when I wrote this comment:

- Basic text: https://gist.github.com/anichitiandreea/e1d466022d772ea22db56399a7af576b
- Emojis: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html
- Red code block background: https://google.github.io/styleguide/tsguide.html

docs/architecture/clients/dependency-injection.md
coroiu: Wow, how did I not get that :D Maybe add "therefore" just in case?

:art: suggestion(non-blocking): 
```suggestion
Non-Angular contexts do not use a DI framework and must therefore manually instantiate their dependencies **in
```

docs/architecture/clients/dependency-injection.md
coroiu: > If the component is standalone, then by definition it doesn't have a module, so advice about a feature module doesn't apply. 

Good point, I read `feature module` as just a folder in an organize-by-feature structure. I think feature modules could (and should) be replaced by standalone component, but that discussion might be outside the scope of this PR comment :D

> Does this go to your earlier concern about accidentally instantiating duplicate services?

Yes, and other things. It depends a little bit on how we define a "feature". If `feature => 1 exported top-level component` then we are basically talking about standalone components. 

In the case where 1 feature module exports multiple components then the module becomes a bit rigid. It's a big blob containing all of the dependencies that any of it's child components *might* need. It makes it less obvious which component needs which service/pipe/etc. and also affects tree-shaking negatively. The lifetime of the service also becomes much easier to reason about, since it's tied to an actual DOM component, instead of a module that is somewhat magically instantiated by Angular in the background. Service tied to components also have access to the activated route in a way that service in modules never will (though this use-case might be a bit of an edge case).

docs/architecture/clients/dependency-injection.md
```diff
@@ -76,6 +79,9 @@ Angular contexts generally use ngModules
 [dependency providers](https://angular.io/guide/dependency-injection-providers) to configure DI. For
 example:
 
+:::danger[DON'T DO THIS] This is a simplified example only - read on to learn about safeProvider.
+:::
+
```
coroiu, line 84: ❌ issue: prettier does this sometimes because it doesn't understand admonitions, to stop it you have to add some new-lines

```suggestion
:::danger[DON'T DO THIS] 

This is a simplified example only - read on to learn about safeProvider.

:::

```

docs/architecture/clients/dependency-injection.md
```diff
@@ -103,6 +109,8 @@ We provide a helper function called `safeProvider`, which acts as a wrapper arou
 entry and provides compile-time type checking of your configuration. For example (continuing with
 the example above):
 
+:::tip[RECOMMENDED] safeProvider is the preferred way to register services in ngModules. :::
+
```
coroiu, line 113: ❌ issue: same here

```suggestion
:::tip[RECOMMENDED] 

`safeProvider` is the preferred way to register services in ngModules. 

:::


```

docs/architecture/clients/dependency-injection.md
coroiu: Keeping this open for future reference! :)

coroiu (commented):
Looking good! One final issue

MGibson1 (changes requested):

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,233 @@
...
+automatically configure its dependencies. If a service is used in mixed contexts (e.g.
+`libs/common`), it must not use Angular decorators and its dependencies must be manually configured.
+
+## Interfaces
+
+A service should implement an interface defined by an abstract class if:
+
+1. The service will be used across multiple client apps, or
+2. You want to define multiple interfaces for your service to limit what functionality is available
+   to different consumers
```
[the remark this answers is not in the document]
eliykat: The purpose of this para isn't to explain TS interfaces vs. abstract classes, which is why I didn't go into that detail. It's solely about your second paragraph.

Jest doesn't require interfaces for mocks, so they're not strictly required for unit testing. I understand the more general point about why interfaces are good for loose coupling, but we don't enforce any such requirement today. A regex all-text search in vscode for `export class [A-Z,a-z]*service \{$` returns 61 services that don't `implement` any abstraction. We can change that but it would be a change.

See also the Slack discussion [here](https://bitwarden.slack.com/archives/C02LPDKPML6/p1712797671412329).

This para is basically trying to come up with some clear rules for why we use abstract classes sometimes but not all the time.

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,233 @@
...
+### Browser
+
+Browser DI is split across the following locations:
+
+- `services.module.ts` is the main services module for the browser extension popup. It imports
+  `JslibServicesModule`
+- `main.background.ts` manually instantiates services used in the background page (for manifest v2)
+  or the service worker (for manifest v3). It does not use any dependency injection framework
+- a series of exported factory functions with the naming convention `[name]-service.factory.ts`.
+  These may be used in manifest v3 service workers in the future
```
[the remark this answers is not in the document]
eliykat: It's used by autofill code today, e.g. https://github.com/bitwarden/clients/blob/main/apps/browser/src/autofill/browser/context-menu-clicked-handler.ts#L118-L138, so it's not just a case of deleting them - we have to replace them with something. That's a whole to-do, do you still think it's safe to be that definitive about it? I mean, either way I agree the current wording isn't really accurate. How about `These are used to a limited extent in autofill code only and will likely be removed in the future.`

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,233 @@
...
+  Angular and non-Angular contexts)**
+- `libs/angular` for services used by **multiple clients** in **Angular contexts only**
+- `apps/<client-name>` for services used by a single client only
+
+Some teams also have their own `libs` folder, which are structured similarly. e.g. `libs/auth`,
+containing `libs/auth/common` and `libs/auth/angular`. These should be used if available.
+
+If a service is used in an Angular context only, it can use the Angular `@Injectable` decorator to
+automatically configure its dependencies. If a service is used in mixed contexts (e.g.
+`libs/common`), it must not use Angular decorators and its dependencies must be manually configured.
```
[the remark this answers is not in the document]
eliykat: It's hard to make simple statements here (this whole page was challenging for this reason). `libs/angular` is an Angular-only context, so "anything in `libs`" is not an accurate example of mixed contexts. Similarly, teams have their own nested Angular libs, e.g. `libs/auth/src/angular`. I've put in your first example though.

docs/architecture/clients/dependency-injection.md
```diff
@@ -0,0 +1,233 @@
...
+- between apps - for example, the web client is entirely Angular, whereas the CLI client does not
+  use Angular at all
+- within apps - for example, the browser client uses Angular in its popup but not in its background
+  page/service worker
+
+This has important consequences for how we write services and how we configure DI. In particular:
+
+- Angular contexts use the Angular dependency injection framework
+- non-Angular contexts do not use any dependency injection framework and must manually instantiate
+  their dependencies
```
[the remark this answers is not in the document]
eliykat: Nice addition!

docs/architecture/clients/dependency-injection.md
eliykat: I added a mention of standalone components when discussing the web vault client, otherwise I suggest this could be done in a subsequent revision of this page - which would definitely be welcome. It might require some discussion/consensus on stricter rules around feature modules though. I agree in many cases they can be replaced with standalone, but it's not what we do today.

docs/architecture/clients/dependency-injection.md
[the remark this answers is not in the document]
MGibson1: They're less used than they appear. Those handlers are invoked with the constructor, not that factory method. that handler factory would be removed as a part of deleting the service factories.

```
These were used early in development for Manifest v3, but will be removed in the near future.
```

docs/architecture/clients/dependency-injection.md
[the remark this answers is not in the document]
MGibson1: good points and discussion, thanks! I agree with your summary, now

docs/architecture/clients/dependency-injection.md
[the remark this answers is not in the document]
MGibson1: nah, that makes sense. The important point is angular only and it's probably best to avoid concrete examples since they're going to be snapshots at best.

MGibson1 (commented):
Responded to comments. Only thing really still outstanding is the factory method statements.

docs/architecture/clients/dependency-injection.md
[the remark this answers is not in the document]
eliykat: Done!

MGibson1 (approved):

PR #343, opened 2024-04-18
SM-1200: Rename service account to machine account

coltonhurst:
## Objective

Rename variants of service account to machine account. This is a bit more complicated for the contributing docs repository, as developers need to know the following:

- "machine account" is synonymous with "service account"
- Front facing text of "service account" is being changed to "machine account"
- We are not renaming service accounts in the code at this time


Thomas-Avery (approved):

MGibson1 (approved):
Honestly once per page would probably have been my choice, but I don't think this is egregious.

PR #345, opened 2024-04-25
[PM-7738] Configure secrets before creating database

mzieniukbw:
## Objective

You need the secrets before you run any database migrator by changing the order of the steps.

### Context

The _Create database_ step describes a step to run `migrate.ps1` command, which in essence runs two commands for MSSQL server:
1. Get the connectionString by running `$(dotnet user-secrets list --json --project ../src/Api | ConvertFrom-Json).'globalSettings:sqlServer:connectionString'`
2. Run the database migrator with the extracted connection string `dotnet run --project ../util/MsSqlMigratorUtility/ "$msSqlConnectionString"`

With the current documentation, a fresh setup would not contain any secrets, since the `secrets.json` file has not been created yet, resulting in some weird connection error, that does not suggest on the actual problem (lack of validation for expected argument being empty in that script is a different problem).

### DIFF
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -181,63 +181,6 @@ the following commands:
     pwsh setup_azurite.ps1
     ```
 
-## Create database
-
-You now have the MSSQL server running in Docker. The next step is to create the database that will
-be used by the Bitwarden server.
-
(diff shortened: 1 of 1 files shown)


Thomas-Avery (approved):
Great catch, thank you

PR #355, opened 2024-05-15
Removing references to the SDK binaries

jhbeskow:
## Objective

We are changing how the Secrets Manager SDK gets pulled into sm-kubernetes.  This change should not be committed until after that change is made.


differsthecat (approved):
Thank you @jhbeskow !

PR #360, opened 2024-05-30
Update used Node and Npm version

djsmith85:
## Objective

With https://github.com/bitwarden/clients/pull/8078 we have bumped Node from V18 to V20. This updates our docs for anyone starting out with their environment.


Hinton (commented):
Can you also upgrade node for building this repo?

djsmith85:
@Hinton Will do in a separate PR, but would prefer to get this information out, as we have merged the bump into `main` on clients.

PR #366, opened 2024-06-10
Run docusaurus locally with SSL if certs are provided

jprusik:
## Objective

Allows devs to quickly/easily generate and provide SSL certs to run `contributing-docs` locally without their https-enforcing browsers getting in the way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,9 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
+    "start": "./scripts/docusaurus-start.sh",
+    "start:insecure": "docusaurus start",
+    "setup:ssl": "./scripts/generate-certs.sh",
     "build": "docusaurus build",
     "swizzle": "docusaurus swizzle",
     "deploy": "docusaurus deploy",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -15,7 +15,9 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
+    "start": "./scripts/docusaurus-start.sh",
+    "start:insecure": "docusaurus start",
```
jprusik, line 19: Just in case folks want/need a way to run the docusaurus command directly

package.json
```diff
@@ -15,7 +15,9 @@
...
+    "setup:ssl": "./scripts/generate-certs.sh",
```
jprusik, line 20: Largely copied over from https://github.com/bitwarden/browser-interactions-testing/blob/main/scripts/generate-certs.sh

scripts/generate-certs.sh
jprusik, line 1: This file is largely copied over from https://github.com/bitwarden/browser-interactions-testing/blob/main/scripts/generate-certs.sh

withinfocus (approved):

PR #398, opened 2024-07-29
Add special note for EDD and EF

rkac-bw:
## 📔 Objective

Add specific info in regards to implementing EDD process for Entity framework Only Databases with Code Migrations (Self host for example)

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/contributing/database-migrations/edd.mdx
+++ b/docs/contributing/database-migrations/edd.mdx
@@ -308,6 +308,26 @@ Prior to merging a PR please ensure that the database changes run well on the cu
 version. We currently do not have an automated test suite for this and it’s up to the developers to
 ensure their database changes run correctly against the currently released version.
 
+## EF-Only Databases with Code Migrations
+
+For EF only databases that use code as migrations (self-host), migrations will run before new code is deployed and therefore before the model that EF works off of has changed, since that is in the code.
+
+Columns that are referenced in queries in the database have to be deleted over 2 releases. Here's the process:
+
+1. Before the first release, delete the property in the C# class.
+
+2. Keep that column in as a shadow property. Shadow properties are properties that are defined only in the EF model but are not defined in the C# classes. This can be particularly useful when you're transitioning columns in the database schema without immediately reflecting those changes in the code.
+
+3. Test creating EF migrations. They should be empty and then can be deleted.
+
+4. Do a release.
+
+5. That column should not be a part of any queries anymore.
+
+6. Delete the shadow property configuration.
+
+7. Create migrations that will drop that column.
+
 ## Further reading
 
 1. [Evolutionary Database Design](https://martinfowler.com/articles/evodb.html) (Particularly


Hinton:
Is this needed? To my knowledge we don't use EF migrations for cloud and unified has no requirement for EDD.

docs/contributing/database-migrations/edd.mdx
```diff
@@ -308,6 +308,26 @@ Prior to merging a PR please ensure that the database changes run well on the cu
 version. We currently do not have an automated test suite for this and it’s up to the developers to
 ensure their database changes run correctly against the currently released version.
 
+## EF-Only Databases with Code Migrations
```
withinfocus, line 311:
```suggestion
## EF-only databases with code migrations
```

withinfocus:
> Is this needed? To my knowledge we don't use EF migrations for cloud and unified has no requirement for EDD.

EF migrations are used for SM in the cloud right? I'd rather be consistent in our approach nonetheless though. It would be ideal to see a PR that includes EDD-prescribed changes both in a traditional DbUp migration script as well as an EF one.

withinfocus (changes requested):
Your lint is failing -- make sure that's running locally for you per [the setup guide](https://github.com/bitwarden/contributing-docs?tab=readme-ov-file#install).

Hinton:
No, SM uses regular migrations since using the EF framework for half our migration and hand written for the rest would be very difficult to understand.

For the delete scenario we should be fine having EF migrations that immediately delete the value. We will lose the ability to roll back though but for a delete that should be fine as we never rollback self-hosted instances.

There are more complex scenarios like column rename that requires some juggling with EF. And requires 3 deployments + 1 data migration.

1. Add column `new`, update code to write both `old` and `new`. <- DB & code deploy.
2. Run DB migration copying all `old` to `new`. <- DB
3. Removing code references to `old`. <- Code
4. Remove `old` from database. <- DB

docs/contributing/database-migrations/edd.mdx
```diff
@@ -308,6 +308,26 @@ Prior to merging a PR please ensure that the database changes run well on the cu
...
+
+For EF only databases that use code as migrations (self-host), migrations will run before new code is deployed and therefore before the model that EF works off of has changed, since that is in the code.
+
+Columns that are referenced in queries in the database have to be deleted over 2 releases. Here's the process:
+
+1. Before the first release, delete the property in the C# class.
+
+2. Keep that column in as a shadow property. Shadow properties are properties that are defined only in the EF model but are not defined in the C# classes. This can be particularly useful when you're transitioning columns in the database schema without immediately reflecting those changes in the code.
+
```
eliykat, line 320: How do I do this? I had a look at https://learn.microsoft.com/en-us/ef/core/modeling/shadow-properties but it wasn't obvious how that maps to our implementation.

eliykat:
Is there a consensus here? I need to drop some columns currently - which approach should I be using?

Regarding the PR itself - it would be useful to have a timeline showing both database frameworks side-by-side. e.g.

---

Example: dropping a column

| C# Entity | MSSQL | EF |
| -- |-- | -- |
| No change | Give the parameter a default value in all stored procedures | No change |
| Release |
| Delete the property (migration) | No change | Configure the column as a shadow property (no migration) |
| Release |
| No change | Drop the column | Delete the shadow property (migration) |

---

Maybe there is a better way to represent this.

eliykat (changes requested):

withinfocus:
As discussed we're got gonna EDD EF at this time.

PR #428, opened 2024-09-16
Authorization deep dive amendments

eliykat:
## 🎟️ Tracking

N/A

## 📔 Objective

Some follow-up changes to the authorization deep dive based on some recent discussions and experience.

The **main change** is to state that `AuthorizationService` should be called from the controller, rather than in the core layer (the query/command class). There are several reasons for this:
1. I still think that authz is fundamentally an api layer concern, mixing it in with business logic in the core layer makes everything more complex. It's useful to think of these separately.
2. Putting it in the command/query limits their re-use. Some queries may be used internally to check state, and are not returned out to the user. Some commands are used as side-effects with fixed argument values that need to be executed for all users, or may be executed by a job that is not tied to a user. Baking authorization into the main command/query logic really confuses things here.
3. It's awkward to inject the `ClaimsPrincipal` into the core layer. On the controller it's always available as `User`.
4. It's more obvious and easier to audit if there's a call in the controller before we enter the core layer.
5. We've already done this and it's been working well.

I have additional ideas around how to handle commands with complex authorization requirements, however I'd rather this guide follow our practice rather than vice versa. I don't want to overspecify here and keep having to amend it. We'll continue to iterate on it.

Other minor changes:
1. Refer to the `Bit.Core.Exceptions.NotFoundException` rather than `NotFoundError` (typo)
2. Don't refer to `AuthorizeOrThrowAsync`. I never added this overload and in retrospect I think it obscures things.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/architecture/deep-dives/authorization.md
+++ b/docs/architecture/deep-dives/authorization.md
@@ -151,12 +151,12 @@ await _cipherRepository.DeleteAsync(cipher);
 Some queries return all resources of a type within a particular scope. For example, rather than
 returning a specific cipher (or a specific set of ciphers), return all ciphers for an organization.
 
-In this case, we use a strongly typed `OrganizationIdResource` wrapper around the organization ID as
+In this case, we use a strongly typed `OrganizationResource` wrapper around the organization ID as
 the resource. The operation describes the scope of the read. This requires a separate handler to be
 defined for this combination of resource and operation.
 
 ```cs
-var authorizationResult = await _authorizationService.AuthorizeAsync(User, new OrganizationIdResource(orgId), CipherOperations.ReadAllForOrganization);
+var authorizationResult = await _authorizationService.AuthorizeAsync(User, new OrganizationResource(orgId), CipherOperations.ReadAllForOrganization);
 if (!authorizationResult.Succeeded)
 {
   throw new NotFoundError();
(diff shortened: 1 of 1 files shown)


withinfocus (commented):
From a guidance perspective I know we have gone back and forth on this one, so is the right answer that it's case-by-case? @mzieniukbw has https://github.com/bitwarden/server/pull/4741 which is pretty clean. Is the callout then to shift the logic when command-chaining comes up? I mostly understand your numbered points but it focuses on being API-centric and that might not be the case for us.

Separately, we should discuss `AuthorizeOrThrowAsync` -- can you elaborate what you think it's obscuring? Is it simply the exception? For something so heavily used I think not repeating the success check is a positive.

eliykat:
The main question here is how tightly coupled authorization and C/Qs should be.

In very simple situations, they can be totally decoupled. e.g. See [SecretsController.CreateAsync](https://github.com/bitwarden/server/blob/f72932bf2496e70776a5fb81f0d4734b3807da9d/src/Api/SecretsManager/Controllers/SecretsController.cs#L99-L105). It's the endpoint for creating a secret, you call `AuthorizationService` with the secret and `SecretOperations.Create`. Pretty straightforward. SM have implemented their entire app this way and we've started to do this in AC Team as well. It means the C/Q can focus on business logic and doesn't need additional mocking in its unit testing.

In more complex situations, the command does multiple things, all of which need to be authorized, and that may not be clear from the controller. In this case I'm going to explore  [an interface](https://github.com/bitwarden/server/blob/6b2aec91184de6e889a11b973648ee246b52c0c4/src/Core/Utilities/Authorization/IGetAuthorizationRequirements.cs) where the command can return a list of its authorization requirements, without making the authz call itself. This is a loose coupling where the controller can still handle authorization without having to know the inner workings of the command; and the command doesn't need to know how authorization works. I want to try it before documenting it but I think this is a better level of abstraction. (It was Gibson's suggestion initially, I can't take too much credit.)

The current guide has the tightest coupling, where the C/Q must call AuthorizationService directly. I think that's undesirable for the reasons in my first post.

I would prefer to have a clear rule about where we do authorization checks, so that they are less likely to be overlooked or missed. (e.g. assuming it's handled by the C/Q class when it's not.) However if it's something that we need to explore more before documenting here then we can do that. It was intended to be a helpful guide for common situations, not a mandatory set of rules.

---

`AuthorizeOrThrowAsync` I feel less strongly about; I can implement it if you think it would be useful.

eliykat:
I have:

1. restored references to `AuthorizeOrThrowAsync` and submitted a PR to add it to the code: https://github.com/bitwarden/server/pull/4790
2. weakened the guidance on where authorization checks should occur so that teams can experiment with what works best for their domain.

PR #439, opened 2024-09-26
Add server setup for GitHub Packages

withinfocus:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-6950

## 📔 Objective

Adds documentation on how to set up GitHub Packages to enable restores of our new NuGet offerings there.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -294,6 +294,29 @@ attempting to "Build and Run the Server".
 
 </Bitwarden>
 
+## Configure NuGet with GitHub Packages
+
+Server-side projects and solutions may use
+[Bitwarden-shared .NET extension libraries hosted on GitHub Packages](https://github.com/orgs/bitwarden/packages?repo_name=dotnet-extensions)
+and these packages require authentication for access.
+
+First, [generate](https://github.com/settings/tokens/new) a GitHub personal access token (classic)
+with the `packages:read` scope only. You can set an expiration date but it may be easier to leave it
+without one considering the scope. Copy the token value and run:
+
+```bash
+IFS= read -rs GITHUB_PAT < /dev/tty
+```
+
+along with pasting the value and pressing Enter. Next, run:
+
+```bash
+dotnet nuget add source --username bitwarden --password $GITHUB_PAT --store-password-in-clear-text --name github --configfile ~/.nuget/NuGet/NuGet.Config "https://nuget.pkg.github.com/bitwarden/index.json"
+```
+
+which will set up the necessary global source and credentials. Any NuGet restores will now also
+utilize our NuGet GitHub Packages.
+
 ## Build and Run the Server
 
 You are now ready to build and run your development server.
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -294,29 +294,6 @@ attempting to "Build and Run the Server".
 
 </Bitwarden>
 
-## Configure NuGet with GitHub Packages
-
-Server-side projects and solutions may use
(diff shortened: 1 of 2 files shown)


justindbaur (commented):
Why not just make the packages public? Then we can add the source to a `nuget.config` file in the repo root? If it's private, and we use the package in one of the server projects community members would be completely blocked from running it.

withinfocus:
Alas this doesn't work, even if they are public -- you have to be authenticated. They are already public, and you can't even pull them without a token being provided. Very limiting.

withinfocus:
I don't love that, at least for now. This is just for us, and it happens to be semi-accessible since we are an open source company. I don't really want people finding these, especially if they search around places like NuGet.

docs/getting-started/server/guide.md
```diff
@@ -294,6 +294,29 @@ attempting to "Build and Run the Server".
...
+without one considering the scope. Copy the token value and run:
+
+```bash
+IFS= read -rs GITHUB_PAT < /dev/tty
+```
+
+along with pasting the value and pressing Enter. Next, run:
+
+```bash
+dotnet nuget add source --username bitwarden --password $GITHUB_PAT --store-password-in-clear-text --name github --configfile ~/.nuget/NuGet/NuGet.Config "https://nuget.pkg.github.com/bitwarden/index.json"
```
justindbaur, line 314: I think we should add a `nuget.config` file with `<add key="github" value="https://nuget.pkg.github.com/bitwarden/index.json" />` in it so that the url is in the repo and doesn't have to be added here. 

Instead the command they run will be: 
```
dotnet nuget update source github --username bitwarden --password $GITHUB_PAT --store-password-in-clear-text --configfile ~/.nuget/NuGet/NuGet.Config
```

This way there will be a more informative error when `dotnet restore` ultimately fails:
```
/usr/share/dotnet/sdk/8.0.101/NuGet.targets(156,5): warning : Your request could not be authenticated by the GitHub Packages service. Please ensure your access token is valid and has the appropriate scopes configured. [/workspace/src/Api/Api.csproj]
/workspace/src/Api/Api.csproj : error NU1301: Failed to retrieve information about 'Bitwarden.Extensions.Hosting' from remote source 'https://nuget.pkg.github.com/bitwarden/download/bitwarden.extensions.hosting/index.json'.
```

Also we will have to add a command that works on windows and we should probably link [here](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#config-file-locations-and-uses) somewhere in the docs. As the location can differ based on installation method.

docs/getting-started/server/guide.md
```diff
@@ -294,6 +294,29 @@ attempting to "Build and Run the Server".
...
 
+## Configure NuGet with GitHub Packages
+
+Server-side projects and solutions may use
+[Bitwarden-shared .NET extension libraries hosted on GitHub Packages](https://github.com/orgs/bitwarden/packages?repo_name=dotnet-extensions)
+and these packages require authentication for access.
+
+First, [generate](https://github.com/settings/tokens/new) a GitHub personal access token (classic)
+with the `packages:read` scope only. You can set an expiration date but it may be easier to leave it
+without one considering the scope. Copy the token value and run:
```
justindbaur, line 305: Probably a good idea to link to here for further reading: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry

justindbaur (commented):
I think I'd still really like if we didn't have to add an additional step to the setup of server, that's something I have been really trying to make more and more simple. 

Nothing bad will happen if someone tries to use the package that isn't us. So as long as we put a disclaimer that only our internal uses are considered when making changes I don't think I feel bad about the slim chance that someone adds this to their own project.

justindbaur (approved):

PR #468, opened 2024-11-04
Ac/pm 14415/add vs code extension cli option

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14415

## 📔 Objective

Add a CLI option to install VS code extensions quickly. This uses the VS code executable, so no extra dependencies are needed.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/getting-started/tools/index.md
+++ b/docs/getting-started/tools/index.md
@@ -142,7 +142,6 @@ code --install-extension ckolkman.vscode-postgres
 
 ```
 
-
 ## Optional tools
 
 The following tools may be useful depending on your preferences or what you’re developing.
--- a/docs/getting-started/tools/index.md
+++ b/docs/getting-started/tools/index.md
@@ -107,6 +107,42 @@ recommended ones include the following:
   - [PostgreSQL](https://marketplace.visualstudio.com/items?itemName=ckolkman.vscode-postgres) -
     syntax highlighting for PostgreSQL
 
+#### CLI Installation
+
+```bash
+# General
+code --install-extension nick-rudenko.back-n-forth
+code --install-extension streetsidesoftware.code-spell-checker
+code --install-extension ms-vsliveshare.vsliveshare
+
+# C#
+code --install-extension ms-dotnettools.csharp
+code --install-extension formulahendry.dotnet-test-explorer
+code --install-extension adrianwilczynski.user-secrets
+
+# Git
+code --install-extension mhutchie.git-graph
+code --install-extension donjayamanne.githistory
+code --install-extension eamodio.gitlens
+
+# TypeScript / Angular
+code --install-extension angular.ng-template
+code --install-extension orta.vscode-jest
+code --install-extension esbenp.prettier-vscode
+code --install-extension dbaeumer.vscode-eslint
+
+# Rust
+code --install-extension rust-lang.rust-analyzer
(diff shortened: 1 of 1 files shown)


JimmyVo16:
I know we’re no longer on BEEEP day, but I thought this would be a good opportunity to practice pushing MRs here, lol. I followed this doc and used Vince’s example [MR](https://github.com/bitwarden/contributing-docs/pull/456) as a guide.
Please let me know if there’s anything I should change. We can also move the feedback to the MR if that’s easier.

withinfocus:
How can these tool installation commands map to the suggestions above this change? I think it would flow better if a CLI command was tied to that. You could also use the URL handler to launch these in the UI too, but the Marketplace site will also offer that so it may be moot.

eliykat:
Thanks for adding to the documentation @JimmyVo16 !

vscode also supports "recommended extensions" for a workspace - see [the current list here](https://github.com/bitwarden/clients/blob/main/clients.code-workspace#L69-L76). Here's how it looks:

Screenshot 2024-11-05 at 9 50 00 AM

I think adding them in the workspace file would make them more discoverable and easier to add, and then you could mention this "recommended extensions" functionality in this document as well.

This can be done as well as or in addition to the CLI commands, I don't mind.

JimmyVo16:
> How can these tool installation commands map to the suggestions above this change? I think it would flow better if a CLI command was tied to that. You could also use the URL handler to launch these in the UI too, but the Marketplace site will also offer that so it may be moot.

I envisioned a use case where a user might want to install all the extensions in one go. Instead of manually clicking through each extension individually, they could simply run this script. Like you mentioned, for cases where someone wants only specific extensions, the current format already accommodates that.

I don't think it's practical to try accounting for more granular cases, like installing half the extensions or specific subsets, because that would introduce significant complexity with the permutations involved.

JimmyVo16:
> Thanks for adding to the documentation @JimmyVo16 !
> 
> vscode also supports "recommended extensions" for a workspace - see [the current list here](https://github.com/bitwarden/clients/blob/main/clients.code-workspace#L69-L76). Here's how it looks:
> 
> Screenshot 2024-11-05 at 9 50 00 AM
> 
> I think adding them in the workspace file would make them more discoverable and easier to add, and then you could mention this "recommended extensions" functionality in this document as well.
> 
> This can be done as well as or in addition to the CLI commands, your choice.

That's a great idea for the future! I believe the current PR brings value as it stands, so we should merge it and potentially revisit this option at a later time.

withinfocus (commented):
I think Thomas has a valid point in just directing this to the recommended extension method, and this is backed up by the [VS Code documentation](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions). As-is this would create a second location to maintain the list, and while the recommendations probably should have been updated in the past it's good to have one place for it, now. Any number of things could have additional setup needs or even error out and one sequential script could obscure that feedback, whereas the UI makes that clear visually. The recommendation method also allows the user to pick and choose what they might want, and I would argue that we should just remove this entire list here in favor of that experience.

JimmyVo16:
@withinfocus 

Those are valid points. I'll close this PR and add the extensions to the workspace.

PR #487, opened 2024-11-25
Add security foundations section

coroiu:
## 📔 Objective

This ports over an internal document while making some adjustments to the principles. Requirements will initially be added as-is

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+sidebar_position: 4
+---
+
+# Requirements
+
+The requirements in this section are organized hierarchically, with top-level requirements defining
+the core rules and obligations that must be met, serving as broad objectives for Bitwarden's
+security model. Sub-requirements expand on these by addressing specific scenarios, exceptions, or
+clarifications, and may override their parent requirement when explicitly stated. Sibling
+requirements at the same level must remain independent and free of contradictions.
```
coroiu, line 11: **thought:** Might be nice to align this to an industry standard like ISO/IEC/IEEE 29148, but there's only so much we can do in a markdown document

docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+     encrypted vault data can be decrypted without any additional information provided by the User.
+     - i. The Client **MAY** store such artifacts if given an _informed and explicit consent_ by the
+       User.
+   - f. The Client **MUST NOT** store any artifacts (e.g. encryption keys) such that the vault data
+     can be decrypted by the Server, regardless of consent.
+
+2. Vault data **MAY** be unprotected while _in use_.
+
+   - a. The Client **MAY** decrypt all vault data during vault unlock.
+     - i. The Client **SHOULD** minimize the quantity of decrypted vault data.
```
coroiu, line 40: thought: we could make this more stringent 

```suggestion
     - i. The Client **SHOULD** limit decrypted data to the records actively in use by the user.
```

docs/security/principles/03-no-security-on-fully-compromised.mdx
```diff
@@ -0,0 +1,21 @@
+# P03 - No security on fully compromised systems
```
coroiu, line 1: **point-of-interest:** this was previously called `No guarantees for unlocked vaults on fully compromised devices`. I changed it to make it more consistent with the other principles, but "no security" might seems a bit harsh

withinfocus:
Thoughts on dropping the acronyms and typed declarations? "P01" and "EK" come off rather strongly to me. Same for the "MUST" and "MUST NOT" all over the place that could probably just be expressed without capitalization and boldface.

withinfocus:
I get where you're coming from with the historical references, but I see those as specifications with a large scope and impact. This is important to the company as principles, but I wouldn't treat it quite so intensely such as to model it vs. writing natural language.

quexten:
Imo:

I prefer the RFC-like MUST NOT, and SHOULD NOT, etc. because it makes it clear that we are unambiguous and very deliberate about whatever meaning we encode with them.
Natural text, even if it's the same words, does not do that. In natural text, "must not" does not necessarily mean an absolute prohibition, and could just be a strong recommendation.

So, if we want to treat this as a set of requirements, that we want to design our applications by, then the RFC-style terms helps make these requirements unambiguous and clear.

coroiu:
My vision was that these requirements would be what we base all of our security design on, and not be "mere guidelines". If that will happen or not is probably beyond the scope of this PR. :)  That said, the primary reason that motivated me to create this documentation was to remove ambiguousness surrounding security design and make it very clear for teams that have to design solutions for our products. Given that I am hesitant to make any changes that could potentially decrease clarity and introduce ambiguousness

docs/security/principles/01-locked-vault-is-secure.mdx
```diff
@@ -0,0 +1,13 @@
+# P01 - A locked vault is secure
+
+Clients must ensure that once the vault has been locked, no vault data can be accessed in plain
+text, even if the device becomes compromised after the lock occurs. Protections are not guaranteed
```
eliykat, line 4: What about vault data that is intentionally stored in plain text? e.g. modification dates, folder guids, boolean values. Or is this excluded by the definition of "vault data"? (is metadata "sensitive and private information"?)

eliykat (commented):
Nice work! A few comments from an AC perspective.

docs/security/principles/01-locked-vault-is-secure.mdx
```diff
@@ -0,0 +1,13 @@
+# P01 - A locked vault is secure
+
+Clients must ensure that once the vault has been locked, no vault data can be accessed in plain
```
eliykat, line 3: How do we reconcile use of obligatory language in this section ("must") with the fact that it's a non-actionable set of guidelines? Is this what the RFC means when it says:

>    Note that the force of these words is modified by the requirement
   level of the document in which they are used.

docs/security/principles/01-locked-vault-is-secure.mdx
```diff
@@ -0,0 +1,13 @@
...
+text, even if the device becomes compromised after the lock occurs. Protections are not guaranteed
+if the device is compromised before the vault is locked.
```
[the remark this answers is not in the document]
coroiu: This is a principle and not a requirement and so not something we would test for directly, that said, you have a point. What about adding a section below on the technical feasibility of actually attaining this goal?

> ## Technical Considerations  
> Achieving this principle depends on the limitations of the platform and environment. For instance, in environments like JavaScript, where memory management is not fully under the control of the client, there may be residual plaintext in memory after the vault is locked. While ideal protection cannot be guaranteed in such cases, efforts must be made to minimize these risks through techniques such as clearing memory buffers and leveraging platform security features when available.

docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+:::warning Draft - Early version
+
+This section is still in its early stages and does not yet reflect current or future standards.
+
+:::
+
+1. Vault data **MUST** be protected _at rest_.
+
+   - a. The Client **MUST** encrypt vault data stored on disk.
+   - b. The Client **MUST** use the UserKey to encrypt vault data stored on the Server.
```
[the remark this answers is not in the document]
coroiu: This part is still a draft and needs work. When it comes to keys I've added a section further down, but I'm going to rely on key management to fill it out

docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+       User.
+   - f. The Client **MUST NOT** store any artifacts (e.g. encryption keys) such that the vault data
+     can be decrypted by the Server, regardless of consent.
+
+2. Vault data **MAY** be unprotected while _in use_.
+
+   - a. The Client **MAY** decrypt all vault data during vault unlock.
+     - i. The Client **SHOULD** minimize the quantity of decrypted vault data.
+   - b. The Client **MAY** leave vault data encrypted after unlock.
+   - c. The Client **MUST** ensure that unprotected data does not linger when no longer _in use_.
```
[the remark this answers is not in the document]
coroiu: Hmmm, I think `is not present in memory` is what I'm looking for here, but that makes this requirement hard to implement. Maybe we should change it to a `SHOULD` at the same time

```suggestion
   - c. The Client **SHOULD** ensure that unprotected data is not present in memory when no longer _in use_.
```

I still think this requirement will have a pretty large impact if adopted since it tends to discourage our huge caches, but having no caches is the current approach in the SDK so hopefully we won't have to prioritize any additional work just for this

docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+This section is still in its early stages and does not yet reflect current or future standards.
+
+:::
+
+1. The UserKey **MUST** have 256 bits of security.
+2. The UserKey **MUST** be protected _at rest_.
+3. The UserKey **MAY** be unprotected while _in use_.
+4. The UserKey **MUST** be protected while _in transit_.
+5. The UserKey **MUST NOT** be shared.
```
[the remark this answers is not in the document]
coroiu: You're actually the second person to give me this exact comment so there is definitely room for improvement here. Account recovery and emergency access is not "sharing" according to the definition: 

> **Data sharing**
> The controlled process in which data leaves the Bitwarden secure environment unprotected. As a consequence the guarantees made by this document will no longer apply. The receiving party may or may not have its own guarantees.

docs/security/principles/01-locked-vault-is-secure.mdx
coroiu: Yes, there is actually still an open thread in confluence about this because I was hoping to get Vault involved in classifying data a bit better. My idea is to split up the vault data into groups of varying sensitivity. In that case we might get:

- High: Password
- Moderate: Username
- Low: Creation date

And then say something like "low sensitivity data" may be stored unprotected

docs/security/principles/01-locked-vault-is-secure.mdx
coroiu: Yeah exactly, principles are guiding stars, they describe general directions and ideals, but in the end it's the requirement that you test against and so that's the language you go after.

When I wrote these I meant to set high standards to encourage us to find new ways of protecting the user. As our technology improves we should be able to use these principles as justification for new and stricter requirements. Because of this I'm hesitant to add loopholes here because some platforms might actually give us the tools necessary to achieve these principles

docs/security/principles/03-no-security-on-fully-compromised.mdx
quexten: Related to the confluence point; I think we might need to add "no security on fully compromised bitwarden contexts"

docs/security/principles/01-locked-vault-is-secure.mdx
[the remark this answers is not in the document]
eliykat: That clarification looks good to me and I think it's useful even though this is only a principle. I suggest that principles should still be broadly achievable even though they don't represent actionable requirements (although maybe I'm still conflating the two a little).

docs/security/requirements.mdx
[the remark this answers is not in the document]
eliykat: I agree with this; being more specific with the requirement while loosening the language a little (MUST -> SHOULD) seems like a good balance given that we currently decrypt everything at once in clients.

docs/security/requirements.mdx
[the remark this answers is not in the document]
eliykat: Interesting. Given that "sharing" has a common use, maybe there is a more specific term that can be used here? (Maybe "export" - or would that be confused with a vault export?)

Alternatively, it's only used a couple of times - maybe you remove the definition and be explicit each time: "The UserKey MUST NOT leave the Bitwarden secure environment in an unprotected state."

---

As a separate question - does this document have anything to say about when data is exchanged/accessed between users then? EA and Account Recovery are within the Bitwarden secure environment, but it has security implications.

docs/security/principles/01-locked-vault-is-secure.mdx
eliykat: I think that would work!

docs/security/requirements.mdx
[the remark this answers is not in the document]
coroiu: Both very good suggestions. I think we might need the definition just to make the document easier to read, but you are right that it might be mixed up with vault export. Still think its worth it though, I'll update it when I get a chance!

docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+## Authentication tokens (AT)
+
+:::warning Draft - Early version
+
+This section is still in its early stages and does not yet reflect current or future standards.
+
+:::
+
+1. The authentication tokens MUST be protected at rest if the client provides a mechanism for secure
```
trmartin4, line 83:
```suggestion
1. The authentication tokens **MUST** be protected at rest if the client provides a mechanism for secure
```

docs/security/requirements.mdx
```diff
@@ -0,0 +1,120 @@
...
+   storage.
+
```
trmartin4, line 85:
```suggestion
2. The authentication tokens **SHOULD** be protected in transit.
```

docs/security/definitions.mdx
```diff
@@ -0,0 +1,118 @@
...
+    - **Secure Notes**: Encrypted notes containing sensitive information that the user wants to keep
+      secure.
+    - **Credit Card Information**: Payment card details like card number, expiration date, CVV, etc.
+    - **Identities**: Personal information such as names, addresses, phone numbers, and email addresses
+      that can be used to autofill forms.
+    - **Attachments**: Any files uploaded by the user to be stored securely within the vault.
+
+  </dd>
+  <dt>User</dt>
+  <dd>The owner of the vault data.</dd>
```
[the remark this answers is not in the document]
coroiu: I tweaked this a little bit to bypass the question of ownership

docs/security/principles/01-locked-vault-is-secure.mdx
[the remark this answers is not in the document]
coroiu: Added the section

docs/security/principles/01-locked-vault-is-secure.mdx
coroiu: I tweak this a little bit by only grouping the data into highly sensitive and less sensitive (see 528330f) to avoid getting into hard discussions. What do you think?

docs/security/requirements.mdx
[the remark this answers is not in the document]
coroiu: Resolving to let KM address this in upcoming PRs

docs/security/requirements.mdx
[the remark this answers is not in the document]
coroiu: Fixed!

docs/security/requirements.mdx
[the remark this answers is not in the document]
coroiu: Renamed sharing -> exporting. Added a new definition for sharing that matches EA

docs/security/requirements.mdx
coroiu: Added! But why SHOULD and not MUST?

docs/security/requirements.mdx
```diff
@@ -0,0 +1,122 @@
...
+:::warning Draft - Early version
+
+This section is still in its early stages and does not yet reflect current or future standards.
+
+:::
+
+1. The authentication tokens **MUST** be protected at rest if the client provides a mechanism for
+   secure storage.
+2. The authentication tokens **SHOULD** be protected in transit.
```
trmartin4, line 86:
```suggestion
2. The authentication tokens **MUST** be protected in transit.
```

trmartin4 (dismissed):
Just one suggestion.

docs/security/requirements.mdx
coroiu: Fixed 👍

trmartin4 (dismissed):

docs/security/principles/05-minimized-impact-of-security-breaches.mdx
```diff
@@ -0,0 +1,11 @@
+# P05 - Minimized impact of security breaches
+
+Even with robust security measures in place, user error or unforeseen vulnerabilities can lead to
+various security breaches, including the compromise of encryption keys or data leaks. Bitwarden
+should take available actions to help users limit the damage caused by such breaches, both in scope
+and duration. This involves:
```
patrick-bitwarden, line 6: Is this intended to be comprehensive? If not, we might say "includes" instead

docs/security/definitions.mdx
```diff
@@ -0,0 +1,135 @@
...
+  <dt>Fully trusted Channel</dt>
+  <dd>
+    A communication channel where all parties have verified each other’s identities. This means the
+    channel provides confidentiality, integrity, and authenticity, ensuring that the data is secure,
+    unaltered, and exchanged only between the trusted parties.
+  </dd>
+  <dt>Data at rest</dt>
+  <dd>
+    Any data that is stored on a device or medium that is not actively being used, processed, or
+    transmitted. This includes (but is not limited to) data stored on disk on the users devices, or on
```
patrick-bitwarden, line 89: "User's"

docs/security/definitions.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+  </dd>
+  <dt>Partially trusted channel</dt>
+  <dd>
+    A communication channel where trust is asymmetrical, meaning only some of the parties trust the
+    channel. One party may have verified the other(s) and thus trusts the channel, while the other
+    party or parties may not have done so, making the channel trusted by one party but untrusted by
+    the other(s).
+  </dd>
+  <dt>Fully trusted Channel</dt>
```
patrick-bitwarden, line 80: "Channel" is capitalized here and not in other headings

docs/security/definitions.mdx
```diff
@@ -0,0 +1,135 @@
...
+message queues, shared memory, and HTTP connections.
+  </dd>
+  <dt>Secure channel</dt>
+  <dd>
+    A communication channel that provides confidentiality and integrity for the data transmitted
+    between two or more parties.
+
+    - **Confidentiality**: The data is unreadable to unauthorized
+    parties, typically using encryption.
+    - **Integrity**: The data cannot be altered or tampered with
```
patrick-bitwarden, line 60: Consider reversing the order of "altered or tampered with" to "tampered with or altered" to avoid the sequence "with without"

docs/security/definitions.mdx
```diff
@@ -0,0 +1,135 @@
...
+    The software application that the User interacts with to access their vault data. This includes
+    the official Bitwarden web vault, desktop applications, mobile applications, browser extensions, and any
+    other software provided by Bitwarden that interacts with the Bitwarden server to access or manage vault data.
+  </dd>
+  <dt>Protected data</dt>
+  <dd>
+    Data stored in a format that is unreadable without additional information. Usually synonymous to
+    encrypted, but with additional expectations about how the key is stored. Encrypted data which is
+    stored together with its decryption key in plain text is not considered to be protected, even
+    thought it is encrypted.
```
patrick-bitwarden, line 43: "thought" should be "though"

docs/security/principles/05-minimized-impact-of-security-breaches.mdx
coroiu: Good point, fixed!

coroiu:
@Hinton I've moved it under a Architecture, though it doesn't really feel like an architectural thing? Either way, I think it's worth it to at least give owners. For the future I wonder if we could move it out again and place it under a new working group's ownership, I'll float that idea in other channels and see where we end up :)

Hinton:
I know, but this at least gives it a home while we figure out exactly where it should live and whom should own it.

withinfocus (approved):
Previously reviewed and I support the other approvals here so far. I think this is a good initial state to land.

patrick-bitwarden (approved):

PR #592, opened 2025-05-21
Remove unused evolutionary database design processes

audreyality:
## 🎟️ Tracking

N/A

## 📔 Objective

Align documentation with present evolutionary database design processes, which manage breaking changes by spanning them across several releases.

[Rendered document](https://remove-unused-edd-processes.contributing-docs.pages.dev/contributing/database-migrations/)

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
  issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #593, opened 2025-05-21
Improve database management documentation

withinfocus:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C065AUDER62/p1747770311208779

## 📔 Objective

Improves documentation given a first pass at #165 as well as some corrections started at #592.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+
+Please take care to ensure:
+
+- any existing stored procedure accepts the same input parameters and that new parameters have
+  nullable defaults
+- when a column is renamed the existing stored procedures first check (coalesce) the new location
+  before falling back to the old location
+- continued updating of the old data columns since in case of a rollback no data should be lost
+
+### Non-backwards compatible migration
```
audreyality, line 67: 🤔 I think we should use affirmative language here. The 'non-backwards compatible migration' is talking about a 'breaking change'.

```suggestion
### Breaking change migration
```

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+:::tip
+
+We recommend reading [T-SQL Code Style](../code-style/sql.md) since it has a major impact in how we
+write migrations.
+
+:::
+
+## SQL database project
+
+The separate database definitions in `src/Sql/.../dbo` serve as a "master" reference for the
```
audreyality, line 23: ⛏️ No need for scare quotes here.

```suggestion
The separate database definitions in `src/Sql/.../dbo` serve as the source of truth for the
```

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+
+1. A backwards-compatible transition migration
+2. A non-backwards-compatible final migration
+
+It is likely that a change does not require a non-backwards-compatible end phase e.g. all changes
+may be backwards-compatible in their final form; in that case, only one phase of changes is
+required. With the use of beta testing, partial roll-outs, [feature flags](../feature-flags.md),
+etc. the often-chosen path is to spread a change across several major releases with a calculated
+future state that can perform a "cleanup" migration that is backwards-compatible but still
+represents an overall-_incompatible_ change beyond the boundaries of what we need for individual
```
audreyality, line 50: ⛏️ Our communication guidelines indicate we should avoid italics and prefer bold because international fonts don't render italics well.

```suggestion
represents an overall-**incompatible** change beyond the boundaries of what we need for individual
```

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+2. Write a migration script, and place it in `util/Migrator/DbScripts`. Each script must be prefixed
+   with the current date.
+
+Please take care to ensure:
+
+- any existing stored procedure accepts the same input parameters and that new parameters have
+  nullable defaults
+- when a column is renamed the existing stored procedures first check (coalesce) the new location
+  before falling back to the old location
+- continued updating of the old data columns since in case of a rollback no data should be lost
```
audreyality, line 65: 👍🏻 Love that you're talking about specific concerns to keep aware of!

docs/contributing/database-migrations/mssql.md
audreyality, line 1: 🤔 Would a simple example of the process make sense? The EDD page has [some sample code](https://contributing.bitwarden.com/contributing/database-migrations/edd#migrations-1); showing where each of those go in the filesystem would be enough, I think.

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+
+In accordance with the tenets of [Evolutionary Database Design](./edd.mdx) every change must be
+considered as split into two parts:
+
+1. A backwards-compatible transition migration
+2. A non-backwards-compatible final migration
+
+It is likely that a change does not require a non-backwards-compatible end phase e.g. all changes
+may be backwards-compatible in their final form; in that case, only one phase of changes is
+required. With the use of beta testing, partial roll-outs, [feature flags](../feature-flags.md),
```
MGibson1, line 47: This is a very twisty sentence.

```suggestion
Most changes are entirely backwards-compatible in their final form. If this is the case, only one phase of changes is required. With the use of beta testing, partial roll-outs, [feature flags](../feature-flags.md),
```

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+etc. the often-chosen path is to spread a change across several major releases with a calculated
+future state that can perform a "cleanup" migration that is backwards-compatible but still
+represents an overall-_incompatible_ change beyond the boundaries of what we need for individual
+release safety.
```
MGibson1, line 51: I don't follow what you're going for with this sentence. If a migration is performed that leaves the DB in an incompatible state, it is by definition a finalization migration for the initial transitional migration. If the DB is always backwards compatible with the `N-1`th state, you can't build up migrations such that it's incompatible with the `N-M`th.

Are you saying that our process is a cleanup so long after transitions that they are effectively unrelated and the finalization paradigm isn't apt?

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Backwards compatible migration
+
+1. Modify the source `.sql` files in `src/Sql/dbo`.
+2. Write a migration script, and place it in `util/Migrator/DbScripts`. Each script must be prefixed
+   with the current date.
+
+Please take care to ensure:
```
MGibson1, line 59: **idea**: Maybe 

```suggestion
Tips to ensure backwards compatibility:
```

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+- continued updating of the old data columns since in case of a rollback no data should be lost
+
+### Non-backwards compatible migration
+
+These changes should be written from the perspective of "all data has been migrated" and any old
+stored procedures that were kept around for backwards compatibility should be removed. Any logic for
+syncing old and new data should also be removed in this step.
+
+1. Remove the backwards compatibility that is no longer needed.
+2. Write a new Migration and place it in `src/Migrator/DbScripts_finalization`. Name it
```
eliykat, line 74: What does "remove the backwards compatibility" in step 1 mean - is it just referring to step 2 or does it require something else (and if so, what)?

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+   `YYYY-0M-FinalizationMigration.sql`.
+   - Typically migrations are designed to be run in sequence. However since the migrations in
+     `DbScripts_finalization` can be run out of order, care must be taken to ensure they remain
+     compatible with the changes to `DbScripts`. In order to achieve this we only keep a single
+     migration, which executes all backwards incompatible schema changes.
```
eliykat, line 79: Does this mean that `DbScripts_finalization` only ever contains 1 file? This is news to me.

Is the finalization script added at the same time as the backwards-compatible change? So there is only 1 finalization script at a time, it is always moved to `DbScripts` after the next rc cut date? What if you need a longer timeframe on your finalization?

docs/contributing/database-migrations/mssql.md
```diff
@@ -0,0 +1,82 @@
...
+
+Upon execution any finalization scripts will be [automatically moved](./edd.mdx#online-environments)
+for proper history.
```
eliykat, line 82: What does "upon execution" refer to - the `_move_finalization_db_scripts` Github workflow?

As mentioned in Slack - how do we handle updating `src/Sql` when moving the finalization files across?

Hinton (changes requested):
We absolutely do need dbo_finalization for any automation to work. `dbo` must at all time accurately represent the state that running all DbScripts results in. The workflow relies on an automation running which moves content from _finalization back just having DbScripts will result in the automation failing due to inconsitencies between dbo and DbScripts.

I've created https://github.com/bitwarden/contributing-docs/pull/594 which rebases this on ps/migrations to make it more clear how they differ.

docs/contributing/database-migrations/mssql.md
Hinton: That was the purpose of `dbo_finalization`.

docs/contributing/database-migrations/mssql.md
Hinton: Yes. It's always been mentioned. 

You break up a pure DB non backwards compatible change into two steps. An example would be the column rename example in https://contributing.bitwarden.com/contributing/database-migrations/edd#migrations-1. 

Initial migration would:
- Add a new column
- Add logic that keeps old and new in sync.

Transition migration would:
- Migrate all data from old to new column.

Finalization migration (run at next deploy):
- Removes the old now unused column.

Hinton: Do note we technically never have to remove the old column but not doing so adds database debt since we have

- Unnecessary data in our database.
- More complex sprocs.

docs/contributing/database-migrations/mssql.md
withinfocus: I support the current term. The point of this documentation is that your change can't go backwards. It may be "breaking" to some, but the intention of the documentation is about direction.

docs/contributing/database-migrations/mssql.md
withinfocus: It's an analogy to a master copy of something, like an album recording. This content just moved its position by the way.

docs/contributing/database-migrations/edd.mdx
audreyality, line 1: 🎨 One of these files should note how our database version policy intersects with [our self-hosted server support policy](https://bitwarden.com/help/bitwarden-software-release-support/#bitwarden-self-hosted-server). Linking from these docs to the help site improves discoverability.

withinfocus: Does it though? My concern is that changes the scope of all this -- everything here is about a single release, forward or backward.

withinfocus:
I addressed a few comments on the smaller things and left responses to the rest. I am gonna move over to `server` and finish the finalization work because there's been a strong reaction to keeping it (and not documenting how things are today, albeit incomplete to reintroduce later), and if that works out can restore the couple sentences about it as well as further bolster the "why".

docs/contributing/database-migrations/mssql.md
audreyality: Yes, I know. "The source of truth" is both more inclusive than "master reference", and more accurate as a description.

docs/contributing/database-migrations/mssql.md
audreyality: It's in the [Technical Writers Style Guide](https://bitwarden.atlassian.net/wiki/x/D4BTN).

docs/contributing/database-migrations/edd.mdx
audreyality: The original scope included this by ensuring finalization can run at any future point after release. If you remove finalization, then you must talk about how we guarantee our self-hosted server maintains its N-2 release compatibility.

withinfocus:
Relates to https://github.com/bitwarden/server/pull/5855 where I'm making the remaining EDD changes.

docs/contributing/database-migrations/mssql.md
withinfocus: Added folder names under each tab for clarity.

withinfocus:
I have no more changes planned for this and open comments are either needing responses or something I am not currently comfortable changing further.

docs/contributing/database-migrations/mssql.md
eliykat: OK gotcha - I've reviewed the existing documentation again and understand the process better now. I think `dbo_finalization` was never automated so that was the source of my confusion.

docs/contributing/database-migrations/mssql.md
eliykat: I could've been clearer in my question - my previous understanding was that you might have several scripts in `DbScripts_finalization` as different teams add different finalization scripts. But this way, you have 1 finalization script at a time, which all teams append to. It's been a while since I read about this process so just trying to get my head around it again.

Hinton (approved):

PR #605, opened 2025-05-30
Introduce ADR-0025: deprecate typescript enums

audreyality:
## 🎟️ Tracking

N/A

## 📔 Objective

Document decision to deprecate typescript enums across the bitwarden/clients codebase.

## 🦮 Reviewer guidelines


docs/architecture/adr/0025-ts-deprecate-enums.md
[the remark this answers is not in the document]
dani-garcia: So I went to look a bit more into what `wasm-bindgen` is doing internally when not using `tsify`, and this is what it generates:

In `*.d.ts` file:
```ts
export enum LogLevel {
  Trace = 0,
   ... all other variants
}
```

In the `*.js` file:
```js
/**
 * @enum {0 | 1 | 2 | 3 | 4}
 */
export const LogLevel = Object.freeze({
  Trace: 0,
  0: "Trace",
  ... all other variants
});
```

Not sure why they have that mapping from the number back to the string value, but other than that this seems fairly similar to what we want (but without the type alias).

I still think we should be encouraging using full discriminated unions with `tsify`, though.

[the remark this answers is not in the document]
audreyality: > Not sure why they have that mapping from the number back to the string value, but other than that this seems fairly similar to what we want (but without the type alias).

It's maintaining compatibility with TypeScript. TS enums support bidirectional mapping. You can use the symbol name to get the value or the value to get the symbol name.

custom-words.txt
```diff
@@ -86,6 +86,7 @@ Yellowpages
 Yubico
 YubiKey
 YubiKeys
+typesafe
```
withinfocus, line 89: ❌ We sort these alphabetically.

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,90 @@
...
+Among them:
+
+- Enums that are not `const` substantially increase output size.
+- Numeric enums implicitly cast from `number` to the enum type, which allows invalid enum values to
+  be transmitted to functions.
+- String enums are named types that require a member is used, which is inconsistent with the
+  behavior of numeric enums.
+
+These inconsistencies cause increased complexity from guard statements in the best of cases. In the
+worst cases, their limitations may be unknown, and thus unguarded. Reducing or eliminating
```
withinfocus, line 24: ❌ Sentence that doesn't end?

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Detection
+
+Typescript deprecation can be linted using a fairly short ESlint plugin. The code has [already been
```
withinfocus, line 28:
```suggestion
TypeScript deprecation can be linted using a fairly short ESLint plugin. The code has [already been
```

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,90 @@
...
+function doSomething(type: CipherType) {}
+
+// And used as a value (just like a regular `enum`)
+doSomething(CipherType.Card);
+```
+
+## Considered Options
+
+- **Allow enums, but advise against them** - This is the current state of affairs. With this option,
+  teams MUST address the readmes, but MAY address them by disabling the lint.
```
withinfocus, line 61:
```suggestion
  teams **must** address the readmes, but _may_ address them by disabling the lint.
```
We do use bold and italic on this site.

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,90 @@
...
+  codebase with typescript objects. Increase the lint to an "error" and prohibit disabling of the
+  lint.
+
+## Decision Outcome
+
+Chosen option: **Deprecate enum use**
+
+### Positive Consequences
+
+- Allows for cases where autogenerated code introduces an enum by necessity
```
withinfocus, line 74:
```suggestion
- Allows for cases where autogenerated code introduces an enum by necessity.
```

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,90 @@
...
+- Developers receive a warning in their IDE, which discourages new enums.
```
withinfocus, line 75:
```suggestion
- Developers receive a warning in their IDE, that discourages new enums.
```

docs/contributing/code-style/enums.md
```diff
@@ -3,7 +3,7 @@
 We don't use Typescript enums because they are not fully type-safe and can cause surprises. Instead
 of using enums, we use constant objects.
 
-## Our Recommended Approach
+## Our Recommended Approach ([ADR-0025](../../architecture/adr/0025-ts-deprecate-enums.md))
```
withinfocus, line 6: 🎨 I didn't dig into the history of this page, but I feel this whole page should be eliminated and the content go to `angular.md`.

docs/architecture/adr/0025-ts-deprecate-enums.md
audreyality: I was aiming for RFC definitions for those words. I guess we don't do that?

docs/contributing/code-style/enums.md
audreyality: I want to take it the other direction. Rather than smoosh together angular and typescript, I want to factor the typescript-specific guidance into a new typescript file. 

As to why:

* Typescript conventions also apply to the CLI and autofill, neither of which use angular.
* There's [angular conventions shadow-documented in the architecture section](https://contributing.bitwarden.com/architecture/clients/presentation/angular#naming-adr-0012) that should be in the angular docs.
* There's also [typescript conventions shadow-documented in the architecture section](https://contributing.bitwarden.com/architecture/adr/typescript-strict#guidelines) that should be in the typescript docs.

That said, this PR focuses solely on the ADR. My intent is to follow up with a separate PR to edit the conventions in more depth.

docs/architecture/adr/0025-ts-deprecate-enums.md
audreyality: Removed in 2b141dd74265953c5565.

audreyality:
📝 I tweaked the reviewers to include the folks that are actively participating in side-discussions.

docs/contributing/code-style/enums.md
withinfocus: Keeping that future enhancement in mind, that works for me.

withinfocus (approved):
I feel this works as a provisional first pass and could be merged as proposed, but up to you if you want to take in more feedback. Said feedback can also come as a result of the council session.

withinfocus (dismissed):

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,196 @@
...
+  teams **must** address the FIXMEs, but _may_ address them by disabling the lint.
+- **Deprecate enum use** - Allow enums to exist for historic or technical purposes, but prohibit the
+  introduction of new ones. Increase the lint to a "warning" and allow the lint to be disabled.
+- **Eliminate enum use** - Prohibit the introduction of any new enum and replace all enums in the
+  codebase with typescript objects. Increase the lint to an "error" and prohibit disabling of the
+  lint.
+
+## Decision Outcome
+
+Chosen option: **Deprecate enum use**
```
coroiu, line 90: issue?: I think the route we took is between the definitions of "Deprecate" and "Eliminate". We made the lint rule throw errors (not warnings) but we also allowed and did disable those errors for existing cases

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Negative Consequences
+
+- Unnecessary usage may persist indefinitely on teams carrying a high tech debt.
+- The lint increased the number of FIXME comments in the code by about 10%.
+
+### Plan
+
+- Update contributing docs with patterns and best practices for enum replacement.
+- Update the reporting level of the lint to "warning".
```
coroiu, line 108: Or are you maybe suggesting a change to our current approach, i.e. to downgrade error -> warning

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -0,0 +1,196 @@
...
+type MappedType = { [K in CipherType]: boolean };
+```
+
+When the compiler examines `instance`, it only knows that the type of each of its members is
+`CipherType`. That is, the type of `instance` to the compiler is
+`{ [K in 1 | 2 | 3 | 4 | 5]?: boolean }`. This doesn't sufficiently overlap with `MappedType`, which
+is looking for `{ [1]: boolean, [2]: boolean, [3]: boolean, [4]: boolean, [5]: boolean }`. The
+failure occurs, because the inferred type can have fewer fields than `MappedType`.
+
+### Workarounds
```
coroiu, line 155: thought: I feel like none of these workarounds actually fix the issue which causes less than possible static type-safety. Are we doing this because we value consistency more than the possibility of statically checked mapped types? Should we consider an "escape hatch" that allows access to the literally typed member in our recommended approach?

```ts
const instance: Record<CipherType, boolean> = {
  [CipherType[Literal].Login]: true,
  [CipherType[Literal].SecureNote]: false,
  [CipherType[Literal].Card]: true,
  [CipherType[Literal].Identity]: true,
  [CipherType[Literal].SshKey]: true,
};
```

I guess this is a perfect topic for the arch meeting :)

docs/architecture/adr/0025-ts-deprecate-enums.md
audreyality: [According to the code](https://github.com/bitwarden/clients/blob/main/libs/eslint/platform/no-enums.mjs#L5-L10), it's presently a suggestion-level lint.

docs/architecture/adr/0025-ts-deprecate-enums.md
```diff
@@ -124,19 +94,105 @@ Chosen option: **Deprecate enum use**
...
 - The warning can direct them to our contributing docs, where they can learn typesafe alternatives.
-- Over time, our code size decreases as enums are replaced.
+- Our compiled code size decreases when enums are replaced.
 - If all teams eliminate enums in practice, the warning can be increased to an error.
 
 ### Negative Consequences
 
 - Unnecessary usage may persist indefinitely on teams carrying a high tech debt.
-- The lint increases the number of FIXME comments in the code by about 10%.
+- The lint increased the number of FIXME comments in the code by about 10%.
```
withinfocus, line 103: ⛏️ This would be present tense; don't understand the change.

withinfocus (approved):
I'll leave final acceptance to the engineers working through the code suggestions here.

withinfocus:
(auto-merge kills me in these situations)

audreyality:
My bad. I figured the remaining suggestions were open questions for the architecture council. 🫠

PR #607, opened 2025-06-02
Recommend same type and value name for enums

coroiu:
Use the same name for the type and the value allows us to emulate regular Enums much better.

[Rendered](https://coroiu-update-const-recommen.contributing-docs.pages.dev/contributing/code-style/enums)

## 📔 Objective

Improve recommendation of const object as enums

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/contributing/code-style/enums.md
+++ b/docs/contributing/code-style/enums.md
@@ -78,7 +78,7 @@ export function asCipherType(value: number): CipherType | undefined {
   return isCipherType(value) ? value : undefined;
 }
 
-export function nameOfCipherType(name: CipherType): keyof CipherType | undefined {
+export function nameOfCipherType(value: CipherType): keyof CipherType | undefined {
   return namesByCipherType.get(value);
 }
 ```


docs/contributing/code-style/enums.md
```diff
@@ -1,25 +1,87 @@
...
+  return namesByCipherType.has(value);
+}
+
+export function asCipherType(value: number): CipherType | undefined {
+  return isCipherType(value) ? value : undefined;
+}
+
+export function nameOfCipherType(name: CipherType): keyof CipherType | undefined {
+  return namesByCipherType.get(value);
+}
```
audreyality, line 83: ⛏️  Just realized I named `name` incorrectly. It's should be `value`.

```suggestion
export function nameOfCipherType(value: CipherType): keyof CipherType | undefined {
  return namesByCipherType.get(value);
}
```

audreyality (approved):
🤔 Do we want to add a utility that goes from a name to an enum value or should that be left as an exercise to the reader?

```
export isCipherTypeName(value: string) : value is keyof CipherType {
    return value in CipherType;
}

export asCipherType(value: number) : CipherType | undefined;
export asCipherType(value: number) : CipherType | undefined;
export asCipherType(value: number | string) : CipherType | undefined {
  if(typeof value === "string") {
    return isCipherType(value) ? value : undefined;
  } else if (typeof value === "number") {
    return isCipherTypeName(value) ? CipherType[value] : undefined;
  }

  return undefined;
}
```

justindbaur (approved):

coroiu:
@audreyality We should probably create more generic enum utils (like you suggested) if we find that these are common use-cases but I don't think we need to start doing it now. I spend half a day trying to make the enum definition code generic and eventually gave up because I didn't want to just rush something that is going to be used by everyone (and also because I don't think it's possible but let's ignore that 🥲 )

PR #641, opened 2025-08-01
ADR for Localization

Hinton:
## 📔 Objective

Proposal for revamping localization by modernizing it and exposing both ICU and interpolation capabilities.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/architecture/adr/0027-angular-i18n-migration.md
+++ b/docs/architecture/adr/0027-angular-i18n-migration.md
@@ -1,0 +1,183 @@
+---
+adr: "0027"
+status: Proposed
+date: 2025-01-08
+tags: [clients, angular, i18n]
+---
+
+# 0027 - Improved Localization
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+## Context and Problem Statement
+
+We currently use a homegrown localization system based on the
+https://developer.chrome.com/docs/extensions/reference/api/i18n api. This has served us reasonably
+well but comes with many limitations which we've had to workaround over the years.
+
+Most notable it lacks any native capability for interpolation which is useful for adding markdown
+such as marking words as italic or bold or inserting links into strings. In most scenarios we've
+handled this by splitting the sentence into multiple localized strings but that is not always
+optimal in all locales.
+
+Another limitation is pluralization where we currently can't distinguish phrases that are singular
+or plural which we've often mitigated by adding (s) at the end which again provides a sub-optimal
+experience as different languages handles names, numbers and dates differently.
+
+In order to better handle more complex localization scenarios we'd like to explore changing the
+underlying localization engine.
+
+## Considered Options
+
+- **Keep chrome.i18n** - Continue using our current homegrown localization system
+- **Build our own** - Create a custom localization framework from scratch
+- **Angular Localization (@angular/localize)** - Use Angular's built-in i18n framework
+- **ngx-translate** - Use the popular Angular translation library
+- **`Transloco`** - Use the modern Angular translation library
+- **angular-i18next** - Use i18next with Angular integration
+
+### Keep `chrome.i18n`
+
+**Pros:**
+
+- We already use it and have existing infrastructure
+- No migration effort required
+
+**Cons:**
+
+- Lacks native interpolation capabilities for HTML
+- No pluralization support
+
+### Build our own
+
+**Pros:**
+
+- Complete control over features and implementation
+- Can be tailored exactly to our needs
+
+**Cons:**
+
+- Development and maintenance overhead
+- Localization is a solved problem with standardized formats
+
+### Angular Localization (`@angular/localize`)
+
+**Pros:**
+
+- Built-in framework for Angular with long-term support
+- Expected to be supported for the lifetime of Angular
+- Seamless integration in Angular environments
+- Advanced capabilities like pluralization and interpolation
+- Uses standardized ICU MessageFormat
+- Provides a command for extracting localized strings from source
+
+**Cons:**
+
+- Supports dynamic localization but documentation is limited
+- Locale must be initialized before bootstrapping Angular (though this shouldn't negatively affect
+  us)
+- Limited support outside Angular - only `$localize` tag available outside templates and only
+  supports basic variable replacement
+- May need pairing with libraries like `intl-messageformat` for content scripts
+
+### `ngx-translate`
+
+**Cons:**
+
+- Angular specific, requires Angular to run
+- Not suitable for content scripts
+
+### `Transloco`
+
+**Cons:**
+
+- Angular specific, requires Angular to run
+- Not suitable for content scripts
+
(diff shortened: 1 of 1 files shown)


Hinton:
@audreyality

> If we're going to couple our i18n directly to our display technology, we'll need a uniform way to communicate user-visible errors up from our shared systems. At present, I'm doing this by communicating the i18n tokens directly, which it seems conflict with the machine-generated identifiers angular uses: "4606963464835766483": "Updated {$ICU}",. (I'm working to eliminate this usage, but it's a thorny problem.)

Yes that's primarily the reason we'd keep the old I18nService around for a transition period since we can't automatically migrate them.

> I'm also a little concerned that this ADR proposes we use multiple i18n solutions to solve the "angular doesn't work inside autofill" problem. What benefit does angular's code present that make the dual-library solution better than just using intl-messageformat ([format.js](https://formatjs.github.io/)) across all JS clients?

That would be the "Build our own" option. I will update the ADR to better reflect that we would most likely improve the existing system than start from scratch.

We can continue expanding our i18n system and add ICU support (through a library). But that only solves half the problem, and we still want HTML tag substitution which means we need something like https://github.com/bitwarden/clients/pull/6302.

Angular provides all this out of the box in a performant manner using the compiler. I don't believe we would be able to build our own pipes and get the same performance, or developer ergonomics.

audreyality:
> While the usages would differ, most developers do not work in both content scripts and angular components. The localized files would all follow the ICU MessageFormat standard but be exposed through the native I18n interface of the chosen framework.

Ah, the standardized format definitely addresses the bulk of my concern. 😄

In addition to autofill, there's the CLI and native clients to think about. We'll want similar if not identical message strings for most features (e.g. field labels, warning messages, etc.) and having a single source of truth for translations across all clients will help cut down on the toil associated with translations.

Is there support for the ICU format on our native clients, too? There is [a project implementing it in Rust](https://docs.rs/icu/latest/icu/), but it's not clear if we're pursuing that.

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+---
+
+# 0027 - Improved Localization
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+## Context and Problem Statement
+
+We currently use a homegrown localization system based on the
+https://developer.chrome.com/docs/extensions/reference/api/i18n api. This has served us reasonably
```
withinfocus, line 15: 🎨 Make into a prettier link and not just show the URL? There are a few of these on this page.

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+well but comes with many limitations which we've had to workaround over the years.
```
withinfocus, line 16:
```suggestion
well but comes with many limitations which we've had to work around over the years.
```

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+
+Most notable it lacks any native capability for interpolation which is useful for adding markdown
```
withinfocus, line 18:
```suggestion
Most notable it lacks any native capability for interpolation which is useful for adding Markdown
```

withinfocus, line 18:
```suggestion
Most notably it lacks any native capability for interpolation which is useful for adding markdown
```
🎨 I wouldn't make this a new paragraph since you're continuing the topic with "most notably".

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+such as marking words as italic or bold or inserting links into strings. In most scenarios we've
+handled this by splitting the sentence into multiple localized strings but that is not always
+optimal in all locales.
+
+Another limitation is pluralization where we currently can't distinguish phrases that are singular
+or plural which we've often mitigated by adding (s) at the end which again provides a sub-optimal
+experience as different languages handles names, numbers and dates differently.
```
withinfocus, line 25:
```suggestion
experience as different languages handles names, numbers, and dates differently.
```

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Angular Localization (`@angular/localize`)
+
+**Pros:**
+
+- Built-in framework for Angular with long-term support
+- Expected to be supported for the lifetime of Angular
+- Seamless integration in Angular environments
+- Advanced capabilities like pluralization and interpolation
+- Uses standardized ICU MessageFormat
```
withinfocus, line 71:
```suggestion
- Uses standardized ICU `MessageFormat`
```
⛏️ A bunch of these technical references could become code.

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+- **Keep chrome.i18n** - Continue using our current homegrown localization system
+- **Continue building our own** - Expand our custom localization framework
+- **Angular Localization (@angular/localize)** - Use Angular's built-in i18n framework
+- **ngx-translate** - Use the popular Angular translation library
+- **`Transloco`** - Use the modern Angular translation library
+- **angular-i18next** - Use i18next with Angular integration
+
+### Keep `chrome.i18n`
+
+**Pros:**
```
withinfocus, line 41: 🎨 I would prefer using `####` for these and dropping the colon.

docs/architecture/adr/0027-angular-i18n-migration.md
```diff
@@ -0,0 +1,185 @@
...
+- May require additional libraries for content scripts
+- Learning curve for development team
+
+### Implementation Plan
+
+Replacing a localization system is difficult since it's used in virtually every component and in
+several background services and content scripts. We therefore need to maintain both systems in
+parallel and ideally offer automated migrations wherever possible.
+
+1. **Proof of concept evaluation**: https://github.com/bitwarden/clients/pull/13737
```
withinfocus, line 171: 🎨 I don't think we need to use boldface in any of these.

docs/architecture/adr/0027-angular-i18n-migration.md
Hinton: This was supposed to be `markup` 🤦

docs/architecture/adr/0027-angular-i18n-migration.md
Hinton: I dropped the colon but didn't make them titles since they would show up in the sidebar which would just be noisy.

withinfocus (approved):
Works for me!

PR #644, opened 2025-08-07
clarify abstraction usage criteria

audreyality:
## 🎟️ Tracking

N/A 

## 📔 Objective

Clarify the criteria and purpose of our abstract class conventions. ([rendered](https://clarify-dependency-injection.contributing-docs.pages.dev/architecture/clients/dependency-injection#interfaces))

## 🦮 Reviewer guidelines

### DIFF
--- a/docs/architecture/clients/dependency-injection.md
+++ b/docs/architecture/clients/dependency-injection.md
@@ -55,8 +55,12 @@ You should have an abstraction for your services:
 2. When your service is exported from a module. All services exported from a module must follow the
    [bridge pattern][bridge-pattern].
 
-For example, you may want to use interfaces to prevent outside modules from calling your internal
-methods:
+If neither of these apply to you, you probably don't need to define an abstract interface for your
+service.
+
+::: tip
+
+You can combine these patterns to create an internal module API:
 
 ```ts
 // Injected in other modules
@@ -75,8 +79,7 @@ class MyService implements InternalMyService {
 }
 ```
 
-If neither of these apply to you, you probably don't need to define an abstract interface for your
-service.
+:::
 
 ## Configuring DI
 
--- a/docs/architecture/clients/dependency-injection.md
+++ b/docs/architecture/clients/dependency-injection.md
@@ -48,11 +48,12 @@ decorators and its dependencies must be manually configured.
 
 ## Interfaces
 
-A service should implement an interface defined by an abstract class if:
+You should have an abstraction for your services:
 
-1. The service will be used across multiple client apps, or
-2. You want to define multiple interfaces for your service to limit what functionality is available
-   to different consumers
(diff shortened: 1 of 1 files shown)


eliykat (dismissed):
Not that I have write access, but

docs/architecture/clients/dependency-injection.md
```diff
@@ -48,34 +48,50 @@ decorators and its dependencies must be manually configured.
...
 
-A service should implement an interface defined by an abstract class if:
+You should have an abstraction for your services:
 
-1. The service will be used across multiple client apps, or
-2. You want to define multiple interfaces for your service to limit what functionality is available
-   to different consumers
+1. When you need to present a subset of your service's API to a consumer (i.e. the [facade
+   pattern][facade-pattern]).
+2. When your service is exported from a module. All services exported from a module must follow the
```
quexten, line 55: So do most angular-only services need an abstraction now by this rule?

audreyality: Whether or not a service is angular-only isn't pertinent to the decision. What matters is whether or not any service crosses a module boundary. That's where the bridge pattern should be used.

I expect that modules that work with angular services (which I take to mean services coupled to angular) wouldn't need to use this pattern in general. They can implement specialized services in the same module as the components that consume them, and thus avoid exporting entirely.

Vault has an [excellent example of this](https://github.com/bitwarden/clients/tree/main/libs/vault/src/services). Their services are almost entirely internal to the `@bitwarden/vault` module. They [export a single abstraction-service pair](https://github.com/bitwarden/clients/blob/main/libs/vault/src/index.ts#L29-L30), and [wire their angular DI internally](https://github.com/bitwarden/clients/blob/main/libs/vault/src/cipher-form/cipher-form.module.ts).

docs/architecture/clients/dependency-injection.md
```diff
@@ -48,34 +48,50 @@ decorators and its dependencies must be manually configured.
...
 
-If neither of these apply to you, you probably don't need to define an abstract interface for your
-service.
+This technique is also useful when breaking changes are necessary and a migration cannot be
+automated. You can create a "v2" abstraction and implement it in the same service as the "v1"
+abstraction, simplifying access to shared data (such as a cache).
+
+The module rule creates programming overhead, but in practice the tradeoff is worth it. The main
+reason for this is that the regularity eases analysis across the codebase. For example, you can
+quickly locate everyone's injectable services with `find -name '*.abstraction.ts' -type f`!
```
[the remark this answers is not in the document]
audreyality: This is descriptive text. It's non-normative, explaining why the pattern is useful and illustrating a common pattern that emerges when you apply the rules.

audreyality:
Closing this PR, since it seems we'll be considering changing this policy instead of refining it.

PR #771, opened 2026-02-19
Clarify upstream merge and Renovate actions on PRs

withinfocus:
## 🎟️ Tracking

Internal discussions after seeing some Renovate activity.

## 📔 Objective

Adds guidance around not continually merging the upstream into Renovate branches given the effects and limitations, as well as in general.


docs/contributing/pull-requests/branching.md
```diff
@@ -106,6 +106,18 @@ It can be helpful to think of this exercise in terms of two dimensions:
   - Introducing changes to one component or page independently others.
   - Introducing changes on a client-by-client basis.
 
+### Keeping your branch up to date
+
+It is usually unnecessary to merge the base branch (e.g. `main`) into a short-lived PR branch.
+Frequent merges from the upstream add noise to the commit history without meaningful benefit; CI
+will surface any conflicts or incompatibilities when the PR is ready to merge. If a rebase is
+needed, prefer rebasing over merging so the branch history stays clean.
```
eliykat: "If a rebase is needed, prefer rebasing over merging" - what is this trying to express? If a rebase is needed then a rebase is needed.

That said, generally our advice among devs is to avoid rebasing, because reviewers lose their history of what commits they've already reviewed. Github has gotten a little better at showing this but merging (rather than rebasing) still avoids the issue altogether. Also, a tidy branch history is nice to have during review, but we squash commit everything to `main` it's less important for us.

docs/contributing/pull-requests/branching.md
```diff
@@ -106,6 +106,18 @@ It can be helpful to think of this exercise in terms of two dimensions:
   - Introducing changes to one component or page independently others.
   - Introducing changes on a client-by-client basis.
 
+### Keeping your branch up to date
+
+It is usually unnecessary to merge the base branch (e.g. `main`) into a short-lived PR branch.
+Frequent merges from the upstream add noise to the commit history without meaningful benefit; CI
+will surface any conflicts or incompatibilities when the PR is ready to merge. If a rebase is
```
eliykat: CI will only surface conflicts or incompatibilities with `main` if you merge `main` into your feature branch so that tests (etc) can run. Is the advice here to update from `main` only immediately before merging, and not before? (I would be OK with that, but it's unclear here and I might be misunderstanding.)

Otherwise it is possible to merge to `main` and then have failures in `main`. e.g. because your database migrations have been overtaken.

eliykat (commented):
I wasn't in the internal discussions, but some input from the tech lead perspective.

docs/contributing/pull-requests/branching.md
eliykat: Rebasing creates the same noise, because you get notifications/CI on the rebased commits, and (imo) merge commits are easy enough to navigate. But perhaps rebasing/force-pushing better represents the semantics of what you're doing (resetting the state of the branch) rather than merges.

Thinking about it more, I'm used to one way of doing it, but realistically we could do either. I would've argued against it before the Github review experience handled it, but I believe it can diff before & after a force push now, so if this is the desired direction then we can work towards it.

For wording, maybe

> If you need to update your branch, prefer rebasing over merging...

withinfocus: It's true that every push fires the same notifications and CI runs regardless of whether the new commit is a merge or a rebase, so "noise" wasn't the right framing. The argument I should have led with is squash-semantics: a rebased branch more closely resembles what main will see after the squash-merge, so the pre-merge state matches the post-merge state.

Your "reviewers lose their history" concern is also real, and the section's actual point is "don't sync often" rather than "rebase vs merge" so I dropped the preference and now present both options neutrally with the tradeoff called out.

docs/contributing/pull-requests/branching.md
withinfocus: The sentence as I wrote it was ambiguous: I was describing post-merge CI on main, not a pre-merge check, and the way I worded it doesn't make that clear. I also added a caveat for the cases that need a pre-merge sync since "CI will catch it after merge" is actively wrong in that scenario.

The intent is to push toward smaller PRs that land quickly so syncing rarely becomes necessary, rather than treating "merge main in weekly" as a default habit.

eliykat (approved):
All my concerns have been addressed, thank you!

PR #793, opened 2026-04-10
[PM-33733] Reorganize SDK documentation

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33733
## 📔 Objective

Refresh of the SDK architecture documentation. Reorganizing it into new sections and expands identified gaps.


docs/architecture/sdk/adding-functionality.md
Hinton, line 1: I wonder if we could make a template using https://crates.io/crates/cargo-generate to streamline this process. Have two templates, one for feature crates and another for util crates.

docs/architecture/sdk/index.md
```diff
@@ -4,16 +4,131 @@ sidebar_position: 3
...
 [Confluence page](https://bitwarden.atlassian.net/wiki/spaces/DEV/pages/517898288/Rust+Learning+Resources).</Bitwarden>
 For API documentation view the latest
 [API documentation](https://sdk-api-docs.bitwarden.com/bitwarden/index.html) that also includes
 internal private items.
 
-[sm]: https://bitwarden.com/products/secrets-manager/
-[pm]: https://bitwarden.com/
+## Architecture Overview
+
+```kroki type=plantuml
```
dani-garcia, line 22: Should we put the new Rust CLI on this diagram as well?

We should also mention utility and foundational crates, and how those generally should not depend on `core`

docs/architecture/sdk/adding-functionality.md
dani-garcia: I've had good luck letting claude do it in the past, but I agree a utility that would create the crate structure, manifest and properly configured client would be nice

docs/architecture/sdk/adding-functionality.md
```diff
@@ -0,0 +1,160 @@
...
+- **Does the functionality require persistent state?** Review the docs for
+  [`bitwarden-state`][state-crate] and see [`bitwarden-vault`][vault-crate] for an example of how
+  state is managed.
+
+- **Does the functionality need the SDK to produce an observable or reactive value?** Migrate the
+  business logic to the SDK and build reactivity on top of it in TypeScript.
+
+## 2. Create the crate
+
+When the functionality warrants its own crate — typically when it represents a distinct domain — add
```
dani-garcia, line 36: Should we add more guidance on whether to create a crate vs reuse an existing one, or how big the scope of a crate should be? `represents a distinct domain` is pretty ambiguous

docs/architecture/sdk/adding-functionality.md
```diff
@@ -0,0 +1,160 @@
...
+#[derive(FromClient)]
+pub struct FoldersClient {
+    pub(crate) key_store: KeyStore<KeyIds>,
+    pub(crate) api_configurations: Arc<ApiConfigurations>,
+    pub(crate) repository: Option<Arc<dyn Repository<Folder>>>,
+}
+```
+
+If the client will be exposed over WASM, annotate it with
+`#[cfg_attr(feature = "wasm", wasm_bindgen)]`.
```
dani-garcia, line 62: We don't mention when creating the crate above that the feature needs to exist. Maybe we just always instruct to add a uniffi/wasm feature to all feature crates?

docs/architecture/sdk/adding-functionality.md
```diff
@@ -0,0 +1,160 @@
...
+
+## 4. Wire into the application interface
+
+Connect the feature client to the SDK `Client` by defining an
+[extension trait](client-patterns.md#extension-traits). This makes the feature accessible without
+modifying `Client` itself.
```
dani-garcia, line 68: Now that we're using the bitwarden-pm and bitwarden-sm clients to define the application, we don't really have as much use for these traits. I wonder if we can skip them entirely and just tell devs to define it directly on `bitwarden-pm` like this?

```rust
impl PasswordManagerClient {
    pub fn vault(&self) -> bitwarden_vault::VaultClient {
        bitwarden_vault::VaultClient::from_client(&self.0)
    }
}
```

docs/architecture/sdk/client-patterns.md
```diff
@@ -0,0 +1,191 @@
...
+If a method needs many or large supporting types (request/response structs, error enums), promote it
+to a subdirectory:
+
+```
+domain/
+├── domain_client.rs         # DomainClient struct definition and initialization
+├── mod.rs
+└── method_name/
+    ├── mod.rs
+    ├── method_name.rs   # impl DomainClient { fn method_name() } + tests
```
dani-garcia: This will cause a clippy warning `clippy::module_inception` if the file has the same name as the parent module, so this is going to be a bit confusing, I think.

We can either move the implementation to the `mod.rs` file or maybe do like we had to do on bitwarden-auth and called the file `method_name_impl.rs` instead.

docs/architecture/sdk/adding-functionality.md
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+```
+
+Consumers access the feature through the application interface:
+
+```rust
+let folders = client.vault().folders().list().await?;
+```
+
```
dani-garcia, line 160: This file has details on how to add functionality up to the `bitwarden-pm` level, we could benefit from some details on how to expose this to UniFFI/wasm too

docs/architecture/sdk/adding-functionality.md
Hinton: It might be useful if another team needs to access your domain since they can do `client.vault()` rather than `VaultClient::from_client(client)`

dani-garcia: Hmm fair point, I wonder if we can just let the macro implement it then

docs/architecture/sdk/adding-functionality.md
Hinton: This ties into my thought for having a template.

docs/architecture/sdk/adding-functionality.md
Hinton: It is, but I feel it's partially up to the teams to identify that.

dani-garcia (dismissed):

dani-garcia (dismissed):

docs/architecture/sdk/index.md
```diff
@@ -4,16 +4,140 @@ sidebar_position: 3
...
 
 <Bitwarden>We have compiled a list of resources for learning Rust in a
 [Confluence page](https://bitwarden.atlassian.net/wiki/spaces/DEV/pages/517898288/Rust+Learning+Resources).</Bitwarden>
 For API documentation view the latest
 [API documentation](https://sdk-api-docs.bitwarden.com/bitwarden/index.html) that also includes
 internal private items.
 
-[sm]: https://bitwarden.com/products/secrets-manager/
-[pm]: https://bitwarden.com/
+## Architecture Overview
```
theMickster: ♻️ The heading case should align with project style guide

```suggestion
## Architecture overview
```

docs/architecture/sdk/index.md
```diff
@@ -4,16 +4,140 @@ sidebar_position: 3
...
+
+[Application Interfaces] -> [Features]
+[Application Interfaces] ---> core
+[Features] --> core
+[Features] ---> Utils
+
+@enduml
+```
+
+## What Belongs in the SDK
```
theMickster: ♻️ The heading case should align with project style guide

```suggestion
## What belongs in the SDK
```

docs/architecture/sdk/index.md
```diff
@@ -4,16 +4,140 @@ sidebar_position: 3
...
+
+**The guiding principle: everything except presentational logic belongs in the SDK.**
+
+The SDK should own all business logic that would otherwise be duplicated across clients. Client code
+should be limited to UI rendering, platform-specific integrations, and calling SDK methods.
+
+### SDK Responsibility
```
theMickster: ♻️ The heading case should align with project style guide

```suggestion
### SDK responsibility
```

docs/architecture/sdk/index.md
```diff
@@ -4,16 +4,140 @@ sidebar_position: 3
...
+| Layer             | Owned By | Examples                                     |
+| ----------------- | -------- | -------------------------------------------- |
+| Presentation      | Client   | UI components, navigation, platform gestures |
+| Business Logic    | **SDK**  | Validation, transformations, calculations    |
+| State Management  | **SDK**  | User state, vault data, sync coordination    |
+| API Communication | **SDK**  | Request/response handling, serialization     |
+| Cryptography      | **SDK**  | Encryption, decryption, key derivation       |
+| Data Models       | **SDK**  | Domain objects, view models                  |
+
+### Decision Checklist
```
theMickster: ♻️ The heading case should align with project style guide

```suggestion
### Decision checklist
```

docs/architecture/sdk/secrets-manager.md
```diff
@@ -1,13 +1,88 @@
...
+Bitwarden SDK designed for external use. The SDK is written in Rust and provides bindings for
+multiple languages. For general binding concepts, see [Language bindings](language-bindings.md). For
+getting started with the Secrets Manager SDK, see the
+[setup guide](/getting-started/sdk/secrets-manager/).
+
+The primary goal of the Secrets Manager SDK is to provide a **stable public API** that shields
+external consumers from the churn of internal SDK APIs. While the internal SDK evolves rapidly to
+support all Bitwarden clients, the SM SDK exposes a curated, backward-compatible surface.
+
+## Architecture Overview
```
theMickster: ♻️ The heading case should align with project style guide

```suggestion
## Architecture overview
```

docs/architecture/sdk/adding-functionality.md
theMickster: A new Claude skill to add to the `ai-plugins` marketplace perhaps 🤔

theMickster (changes requested):
The PR is well-written and does a great job to tidy up a number of heading case issues that resided from prior work. A couple minor adjustments to new headers and I think we are in great shape. Thanks for the great work here, @Hinton!

theMickster (approved):

docs/architecture/sdk/index.md
```diff
@@ -4,16 +4,140 @@ sidebar_position: 3
...
+across all Bitwarden clients. It serves as the single source of truth for core business logic.
+Written in Rust, the SDK is versatile and provides bindings for a variety of platforms, including
+mobile clients (Kotlin and Swift) and web clients (JavaScript/TypeScript). The general aspiration is
+to write as much code as possible once in the SDK and have it consumed by all clients, ensuring
+feature parity and reducing duplication.
 
 <Bitwarden>We have compiled a list of resources for learning Rust in a
 [Confluence page](https://bitwarden.atlassian.net/wiki/spaces/DEV/pages/517898288/Rust+Learning+Resources).</Bitwarden>
 For API documentation view the latest
 [API documentation](https://sdk-api-docs.bitwarden.com/bitwarden/index.html) that also includes
```
mzieniukbw, line 17: This link is broken
```suggestion
[API documentation](https://sdk-api-docs.bitwarden.com/index.html) that also includes
```

PR #798, opened 2026-04-17
Updated user defined types section

mkincaid-bw:
## 📔 Objective

Updated "User defined types" section to clarify our stance on new user-defined types.

### DIFF
--- a/docs/contributing/code-style/sql.md
+++ b/docs/contributing/code-style/sql.md
@@ -548,13 +548,6 @@ table-valued parameters in stored procedures:
 - **`[dbo].[EmailArray]`** — a single-column table of `NVARCHAR(256)` email addresses. Use when
   passing a list of emails to a stored procedure.
 
-- **`[dbo].[CollectionAccessSelectionType]`** — a table representing collection access grants with
-  columns `Id` (GUID), `ReadOnly` (bit), `HidePasswords` (bit), and `Manage` (bit). Use when
-  assigning or updating collection access for multiple users or groups in a single operation.
-
-- **`[dbo].[OrganizationSponsorshipType]`** — a table representing full `OrganizationSponsorship`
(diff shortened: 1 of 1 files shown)


withinfocus (changes requested):
Can you expand on what those existing types are a bit? Namely the GUID array one. That may not be appropriate here too, and so perhaps this whole section is removed.

mkincaid-bw:
> Can you expand on what those existing types are a bit? Namely the GUID array one. That may not be appropriate here too, and so perhaps this whole section is removed.

Sure, I expanded on the existing types. I think it may be helpful but if you prefer I remove the whole section, let me know.

rkac-bw (dismissed):
lgtm

docs/contributing/code-style/sql.md
```diff
@@ -536,15 +536,24 @@ WHERE
...
+
+- **`[dbo].[EmailArray]`** — a single-column table of `NVARCHAR(256)` email addresses. Use when
+  passing a list of emails to a stored procedure.
+
+- **`[dbo].[CollectionAccessSelectionType]`** — a table representing collection access grants with
+  columns `Id` (GUID), `ReadOnly` (bit), `HidePasswords` (bit), and `Manage` (bit). Use when
+  assigning or updating collection access for multiple users or groups in a single operation.
+
+- **`[dbo].[OrganizationSponsorshipType]`** — a table representing full `OrganizationSponsorship`
+  records. Use for bulk insert or update of sponsorship rows.
```
withinfocus: 💭 These are really specific and are examples of types that should probably be removed in my opinion, but we're here now. Do we still document them?

mkincaid-bw: Good point. I'll remove them from the docs.

withinfocus (approved):
Thanks for bearing with me!

PR #802, opened 2026-04-28
Native Messaging Test Runner - PR Reviews Update

coltonhurst:
## 🎟️ Tracking

None

## 📔 Objective

Small update to the section about expected PR Review testing for breaking dep changes for the native messaging test runner.

### DIFF
--- a/docs/getting-started/clients/desktop/native-messaging-test-runner.md
+++ b/docs/getting-started/clients/desktop/native-messaging-test-runner.md
@@ -136,6 +136,8 @@ guidelines for what to validate:
   that are impacted.
   - For example: If there are cipher changes, the `status` and `bw-credential-create` should be
     tested.
+- If the PR is a breaking change dependency update, `status` and one of the create / read / update
+  commands should be tested (for example, `bw-credential-create`).
 - If the PR is a non-breaking change dependency update, `status` is sufficient.
 
 ## Troubleshooting


docs/getting-started/clients/desktop/native-messaging-test-runner.md
```diff
@@ -136,6 +136,8 @@ guidelines for what to validate:
   that are impacted.
   - For example: If there are cipher changes, the `status` and `bw-credential-create` should be
     tested.
+- If the PR is a breaking change dependency update, `status` and one of the create / read / update
+  commands should be tested (for example, `bw-credential-create`).
```
differsthecat, line 140: How does one know if it is a breaking dependency update vs non-breaking?

coltonhurst: Ideally this should be in the update notes from the package maintainer. For example, with the [yargs v18](https://github.com/bitwarden/clients/pull/18718) update, in the update notes they included how it was a breaking change. Additionally, we hopefully should be able to tell from the [semver](https://semver.org) update.

However, in theory, breaking changes could be introduced and the maintainer not make it clear through their notes or semver. If we want to go the safest route, then we can test `status` and one of the create / update / read commands on every dependency update, or test all commands on every update.

neuronull: I think it'd be useful to add some detail in this new content that helps clarify that, and I agree with your statements: (semver ideally but not all projects follow it -> the dependency management bot might flag it but also cant rely on that always)

the only addition I may add is that a "clean" dep update to me looks like no manual changes on our part. If we have to modify the source code or manually edit config files etc. , that to me warrants additional testing beyond the minimums.

coltonhurst: Yeah, those are fair points. Thinking a bit more about this, it feels like a separate section for dependency management is warranted, as these semver / breaking changes discussions and how they impact testing are most likely consistent across most dependencies we have. Then, teams could update relevant sections (like this) on how breaking changes impact their dependency areas.

Any thoughts? If we feel like that's accurate, then I'll make a ticket to create that page, and consider that separate from this.

neuronull: It sounds like you're suggesting adding a new docs page (that isn't specific to Desktop), and then we reference that here? If so, it feels like something broader audiences would want to weigh in on (but I might be misunderstanding). But maybe @differsthecat  has advice on this!

differsthecat: Dependency management is more of an internal concern than an external one, so I'd advocate keeping it out of the contributing docs. With that, we may not need to add anything additional to this point in the docs, and keep it in internal process docs.

differsthecat (approved):

PR #803, opened 2026-05-04
Update guidance on licensing cert

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36250

## 📔 Objective

The current contributing docs prescribe loading the licensing certificate as a root certificate into the OS keychain. This gives everyone with the root certificate - all bitwarden employees - the ability to spoof, intercept and tamper with all TLS traffic of all everyone who has this trusted root certificate in their keychain.

Instead of loading it into the keychain as a root certificate, this PR adds the ability to load it from a file, only for the specific scope of the `bitwarden/server` application, entirely eliminating this risk.

### DIFF
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -306,8 +306,17 @@ To run your local server environment as a licensed instance, you will need to do
 1. Log in to your company-issued Bitwarden account
 2. On the "Vaults" page, scroll down to the "Licensing Certificate - Dev" item
 3. View attachments and download both files
-4. Create a `~/.secrets/` folder and place the dev.cer inside
-5. Add the path to the `dev.cer` to `secrets.json` under `LicenseCertificatePath`
+4. Create a `~/.secrets/` folder and place `dev.pfx` inside
+5. Add the following under `globalSettings` in `secrets.json`, substituting your own path to
+   `dev.pfx` and the password from the "Licensing Certificate - Dev" vault item:
+
+   ```json
+   "licenseCertificatePath": "/Users/<your name>/.secrets/dev.pfx",
+   "licenseCertificatePassword": "<password from the vault item>"
+   ```
+
+6. Re-run `setup_secrets.ps1` to apply the new values to every server project — see
+   [Configure user secrets](#configure-user-secrets).
 
 :::warning
 
--- a/docs/getting-started/server/guide.md
+++ b/docs/getting-started/server/guide.md
@@ -309,8 +309,12 @@ To run your local server environment as a licensed instance, you will need to do
 4. Create a `~/.secrets/` folder and place the dev.cer inside
 5. Add the path to the `dev.cer` to `secrets.json` under `LicenseCertificatePath`
 
-:::warning Do not import the "Licensing Certificate - Dev" into your keychain. Doing so may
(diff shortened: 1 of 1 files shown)


docs/getting-started/server/guide.md
```diff
@@ -301,23 +301,20 @@ up-to-date. See [MSSQL Database](./database/mssql/index.md) for more information
...
+`Licensing Certificate - Dev` from the shared Engineering collection.
 
 1. Log in to your company-issued Bitwarden account
 2. On the "Vaults" page, scroll down to the "Licensing Certificate - Dev" item
 3. View attachments and download both files
-4. Go to Keychain Access and set the dev.cer certificate to "Always Trust"
-5. The dev.pfx file will ask for a password. You can get this by clicking and opening the Licensing
-   Certificate - Dev item in the vault
+4. Create a `~/.secrets/` folder and place the dev.cer inside
+5. Add the path to the `dev.cer` to `secrets.json` under `LicenseCertificatePath`
```
Thomas-Avery: It's unclear that `LicenseCertificatePath` is under globalSettings. I would suggest making an example JSON block and direct them to put it in `globalSettings` like we do here https://contributing.bitwarden.com/getting-started/server/advanced-setup#file-uploads-file-sends-and-attachments. 

I would also suggest adding a call to action that configuring the user secrets needs to be re ran https://contributing.bitwarden.com/getting-started/server/guide#configure-user-secrets.

Thomas-Avery: ❓ I don't see anything about the other entry `LicenseCertificatePassword`? Without putting anything in `secrets.json` this will fail to load the cert.

Thomas-Avery (changes requested):

quexten:
@Thomas-Avery Applied your feedback. Both good points!

docs/getting-started/server/guide.md
```diff
@@ -301,23 +301,29 @@ up-to-date. See [MSSQL Database](./database/mssql/index.md) for more information
...
-Mac users: When prompted to save the downloaded certificate and PFX file in Keychain Access be sure
-to select "Default Keychain > login" from the dropdown otherwise they will not be found when
-attempting to "Build and Run the Server".
-
-:::
+`Licensing Certificate - Dev` from the shared Engineering collection.
 
 1. Log in to your company-issued Bitwarden account
 2. On the "Vaults" page, scroll down to the "Licensing Certificate - Dev" item
 3. View attachments and download both files
```
theMickster: ⚠️ Step 3 instructs an engineer to download _both files_ but only `dev.pfx` is used in the new flow.  A new engineer following these instructions will download `dev.cer` but find no instruction for it. 

```suggestion
3. View attachments and only download `dev.pfx`
```

theMickster (changes requested):
Minor change that was caught by a [local multi-agent](https://github.com/bitwarden/ai-plugins/pull/96) Claude Code session.

docs/getting-started/server/guide.md
theMickster: Applied in 593d2c38fb8fced2ccdfd9d01cb70b4da3421763

theMickster (approved):

Thomas-Avery (approved):
LGTM

PR #672, opened 2025-09-17
Update commit signing docs to include how to avoid macOs passphrase prompt per commit + misc refactors

JaredSnider-Bitwarden:
## 🎟️ Tracking

n/a

## 📔 Objective

See title.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
  issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #805, opened 2026-05-05
docs: restore hardware key callout to top-level SSH signing tip

addisonbeck:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/GRN5AAZ6Y/p1778000211187909

## 📔 Objective

On https://github.com/bitwarden/contributing-docs/pull/672 we moved the sk-backed key recommendation out of the main tip and into a nested step. Add it back at the top level so hardware keys remain visibly ranked as the highest-security option.


eliykat (approved):

djsmith85 (commented):
FYI: Vince has some additional changes here: https://github.com/bitwarden/contributing-docs/pull/806

PR #806, opened 2026-05-06
Update Commit Signing Doc

vgrassia:

### DIFF
--- a/docs/contributing/commit-signing.mdx
+++ b/docs/contributing/commit-signing.mdx
@@ -2,38 +2,179 @@
 sidebar_position: 3
 ---
 
-import Tabs from "@theme/Tabs";
-import TabItem from "@theme/TabItem";
-
 # Commit signing
 
-It's possible to configure git with any name and email, enabling bad actors to spoof commits and
-impersonate whomever they want. GitHub supports several ways to digitally sign git commits,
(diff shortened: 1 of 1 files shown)


docs/contributing/commit-signing.mdx
```diff
@@ -2,337 +2,379 @@
...
-(YubiKey or other FIDO2 device) as your signing key.
-
-:::
+There are three ways to set up SSH commit signing, in order of recommendation:
 
-### Standard SSH key configuration
+1. **[Hardware-backed SSH key](#hardware-backed-ssh-key)**: private key lives on a FIDO2 security
+   key (e.g., YubiKey 5+). Each commit requires a PIN and a physical touch. This is the most secure
+   option.
+2. **[Standard SSH key](#standard-ssh-key)**: private key lives on disk, protected by a passphrase.
```
quexten, line 24: Why are we recommending on-disk SSH keys over bitwarden ssh agent? On most systems, the standard SSH key has fewer protections (no memory-dump protection, possibly no passphrase, and also no access control), wheras the Bitwarden-stored SSH agent has all three (assuming you did not configure to never ask for approval)?

quexten (changes requested):

PR #816, opened 2026-06-11
[PM-38553] Key Connector on .NET 10

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38553

## 📔 Objective

Update Key Connector to .NET 10:
- Changed `.NET 8.0 SDK` to `.NET 10.0 SDK`
- Removed the entire "macOS" section - it was outdated, now .NET 10 ships on arm64 arch and does not require openssl steps
- Removed the macOs ARM dotnet path note in the "Running and Debugging" -> "CLI" tab - not necessary, since setup now uniformly says .NET 10 SDK
- Updated the Key Connector URL - since it default to `<web-vault-url>/key-connector` in web vault, when configuring org SSO with Key Connector url.


Thomas-Avery (approved):
LGTM

PR #818, opened 2026-06-15
[PM-39035] Document client lock states

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39035

## 📔 Objective

Add a deep-dive page documenting the lock states a Bitwarden client moves through: logged out, locked before-first-unlock (BFU), locked after-first-unlock (AFU), and unlocked. Explains that some secrets are held in memory only and re-hydrated on unlock, so feature behavior differs by state.

## 📸 Screenshots

n/a (docs only)


withinfocus (dismissed):
Significant but clean documentation improvements. You can wait on someone from your team to also give the nod on a deeper technical review but this works for me.

docs/architecture/deep-dives/lock-states.md
```diff
@@ -0,0 +1,63 @@
...
+(see [Memory hardening](./memory-hardening.md)). As a result, the same feature can behave
+differently depending on the current state.
+
+## States
+
+| State            | Description                                                                                                                                                                                                                              |
+| ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Logged out**   | No account is loaded. No keys or secrets are present.                                                                                                                                                                                    |
+| **Locked (BFU)** | _Before First Unlock._ An account is loaded but has not been unlocked since the process started (fresh launch, after a restart, or after a process reload). Only data persisted at rest is available; in-memory-only secrets are absent. |
```
Thomas-Avery: Typically in technical writing we introduce the full term and follow it up immediately with the acronym in parentheses. Subsequent use is just the acronym in the document. 




```suggestion
| **Locked - Before First Unlock (BFU)** | An account is loaded but has not been unlocked since the process started (fresh launch, after a restart, or after a process reload). Only data persisted at rest is available; in-memory-only secrets are absent. |
```

docs/architecture/deep-dives/lock-states.md
```diff
@@ -0,0 +1,63 @@
...
+| **Locked (AFU)** | _After First Unlock._ The vault was unlocked at least once during this session and then locked. In-memory-only secrets derived during the first unlock are still retained in protected memory.                                           |
```
Thomas-Avery: Same here


```suggestion
| **Locked - After First Unlock (AFU)** | The vault was unlocked at least once during this session and then locked. In-memory-only secrets derived during the first unlock are still retained in protected memory.                                           |
```

docs/architecture/deep-dives/lock-states.md
```diff
@@ -0,0 +1,63 @@
...
+Several features depend on secrets that are only available after the first unlock. In the BFU state
+those secrets are absent, so the feature is either unavailable or behaves differently.
+
+### PIN unlock
+
+When PIN unlock is configured to require a master password reprompt, the material needed to satisfy
+the reprompt is only retained in memory after the first unlock. In the AFU state the PIN can be used
+to unlock directly. In the BFU state that material is not present, so the user must fall back to a
+full master password unlock. To prevent the application from going back into the BFU state, PIN
+unlock may prevent process reload from occurring.
```
Thomas-Avery: > To prevent the application from going back into the BFU state, PIN
unlock may prevent process reload from occurring.

The may here leaves things unclear. Something like?

"when PIN unlock is configured in AFU mode, this prevents process reload from occurring"

Thomas-Avery (changes requested):
Looks good some minor things from me.

docs/architecture/deep-dives/lock-states.md
```diff
@@ -0,0 +1,63 @@
...
+
+Several features depend on secrets that are only available after the first unlock. In the BFU state
+those secrets are absent, so the feature is either unavailable or behaves differently.
+
+### PIN unlock
+
+When PIN unlock is configured to require a master password reprompt, the material needed to satisfy
+the reprompt is only retained in memory after the first unlock. In the AFU state the PIN can be used
+to unlock directly. In the BFU state that material is not present, so the user must fall back to a
+full master password unlock. When PIN unlock is configured in AFU mode, this prevents process reload
```
theMickster, line 34: ♻️/🎨 Safe to assume AFU mode is AFU state, correct? If so then likely best to be 100% precise since this document is a deep-dive. 

```suggestion
full master password unlock. When PIN unlock is configured in AFU state, this prevents process reload
```

theMickster (commented):
One point of clarification; non-blocking IMO but worth a confirmation.

Thomas-Avery (approved):
LGTM

docs/architecture/deep-dives/lock-states.md
quexten: Actually, mode is correct here. The client can be in the AFU state (not mode), but pin unlock is configured in AFU mode, meaning "pin unlock is available when the client is in the AFU state, but not in the BFU state" (mapping to only storing the pin envelope in memory, not on disk.

https://github.com/bitwarden/sdk-internal/blob/61b6ed044280fe49f6eadc69c8031a9972b52be4/crates/bitwarden-core/src/key_management/pin_lock_system.rs#L38
https://github.com/bitwarden/sdk-internal/blob/61b6ed044280fe49f6eadc69c8031a9972b52be4/crates/bitwarden-core/src/key_management/pin_lock_system.rs#L266

theMickster: Okay, perhaps we consider additional clarifications now in a follow-up PR. 
If you chose not to then that's cool too, it's just a suggestion. 

My line of thinking was...
Me (coming in as an outsider trying to understand the deep-dive for the first time) reads the word _state_ 15 times throughout the document, and in this one sentence we introduce _mode_ which made me 🛑 + 🤔  I tried to think like a first-time reader would if I wasn't 100% sure on terminology... I did my own digging through our Contributing Docs and [our Help Center](https://bitwarden.com/help) for the differences between AFU state and AFU mode and found none 🤷🏼‍♂️

theMickster (approved):

PR #820, opened 2026-06-19
Add autoconfirm policy as an exception to zero knowledge

eliykat:
## 🎟️ Tracking

Internal change

## 📔 Objective

Update _Security Principle 01: Servers are zero knowledge_ to note the auto-confirm policy as an exception. This is already noted in our [help documentation](https://bitwarden.com/help/automatic-confirmation/#potential-security-risks), but it's also relevant here because this says that it's a complete list of all exceptions.


theMickster (approved):

quexten (commented):
I realize this is already merged, so no review needed, but very nice work. THank you for adding this @eliykat !!!

PR #825, opened 2026-06-24
Add ADR 31 about minimal API adoption

justindbaur:
## 🎟️ Tracking

Relates to: [ARCH-16](https://bitwarden.atlassian.net/browse/ARCH-16)

## 📔 Objective

Adds an ADR proposing that all new server endpoints be built using .NET minimal APIs in feature-scoped libraries, rather than continuing to add to the existing controller-based approach.

  The ADR covers:
  - Why now: .NET 10 closes the last significant feature gap (automatic validation), and we've already proven a smaller slices architecture in clients and sdk
  - The required library shape: `Add[Feature]Services` + `Map[Feature]Endpoints` with a code example showing side-by-side usage with existing controllers
  - Migration strategy: existing controllers migrated opportunistically, with a more aggressive timeline for endpoint-light projects (Icons, Notifications, Events)
  - Trade-offs: mixed codebase during transition, and `ActionFilterAttributes` needing rewrite before affected controllers can be migrated

## 📸 Screenshots

[ARCH-16]: https://bitwarden.atlassian.net/browse/ARCH-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #826, opened 2026-06-25
Add ADR 32 about breaking up Core project

justindbaur:
## 🎟️ Tracking

Relates to: [ARCH-16](https://bitwarden.atlassian.net/browse/ARCH-16)

## 📔 Objective

Adds an ADR proposing that the `Core` project should be decomposed into smaller feature scoped libraries. 

In draft until #825 merges.

## 📸 Screenshots

[ARCH-16]: https://bitwarden.atlassian.net/browse/ARCH-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/architecture/adr/0032-break-up-core.md
+++ b/docs/architecture/adr/0032-break-up-core.md
@@ -1,0 +1,193 @@
+---
+adr: "0032"
+status: "Proposed"
+date: 2026-06-25
+tags: [server]
+---
+
+# 0032 - Break up the Core project
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+## Context and problem statement
+
+The `Core` project in the `server` repository has grown into a catch-all library that almost every
+service depends on. This creates two significant problems:
+
+- **Unowned code and unbounded dependencies** — while much of `Core` is owned by specific teams, a
+  significant portion has accumulated without clear ownership, making it a dumping ground for shared
+  utilities and dependencies that don't have a better home.
+- **Limited independent deployability** — because every service depends on `Core`, any change to
+  `Core` technically constitutes a change to every service. This makes it hard to reason about the
+  blast radius of a change and undermines the ability to deploy services independently with
+  confidence.
+
+`GlobalSettings` compounds these problems. It is a single configuration class that houses settings
+for every feature and service, meaning all of `Core`'s consumers must load and be aware of the
+entire settings surface even when they only need a small slice of it. As features are extracted from
+`Core`, they should define their own strongly-typed options classes rather than growing
+`GlobalSettings` further.
+
+Not all of what currently lives in `Core` needs to move into feature-scoped libraries. A number of
+cross-cutting concerns — feature flag evaluation, version info endpoints, security middleware, and
+caching — are being built into the server SDK. As those SDK packages mature, the corresponding code
+in `Core` can be removed and replaced with an SDK dependency, further shrinking `Core`'s footprint.
+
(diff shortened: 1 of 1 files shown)


docs/architecture/adr/0032-break-up-core.md
```diff
@@ -0,0 +1,193 @@
...
+  Libraries/
+    Mailer/       # settings, services, repositories, endpoints for the Mailer feature
+    Push/
+    Vault/
+    ...
+  Services/
+    Api/          # composes libraries into a deployable service
+    Identity/
+    Notifications/
+    ...
```
Hinton, line 79: Can we consider how bitwarden licenses libraries and services will look? I think we will start to see more of those. Are they under a different top namespace, or do libraries and services follow a prefix? (Currently libraries loosely use a `Commercial` prefix.

justindbaur: It should look the exact same as this layout just starting at `bitwarden_license`. I don't really see a need for a different namespace. If things are feature scoped I don't think things should collide.

Hinton: Probably. I think it's worth including an example or two though.

justindbaur: Done

theMickster: nice. 
addressed in 9597c125ec29f1c629d29e999e085c2275e80acc

docs/architecture/adr/0032-break-up-core.md
```diff
@@ -0,0 +1,206 @@
...
+  }
+  ```
+
+  **After:**
+
+  The feature library declares its options type and consumes it via `IOptions<T>`. It does not bind
+  configuration itself — that is the host service's responsibility, since the host owns its service
+  settings and knows which configuration sections map to which feature options.
+
+  ```csharp
```
theMickster, line 172: ♻️ Given the guidance we just codified in [0031-adopt-minimal-apis.md#decision-outcome](https://github.com/bitwarden/contributing-docs/blob/main/docs/architecture/adr/0031-adopt-minimal-apis.md#decision-outcome), I think we consider renaming the two locations where we call our `AddMailers` to instead be `AddMailerServices`.

-OR- if we are going to not rename the extension method during a migration, then we should consider changing the examples to `AddMailer` because that's the actual name of the extension method. 

See these for more details on my finding.
- `src/Core/Platform/Mail/README.md`
- `src/Core/Platform/Mail/Mailer/MailerServiceCollectionExtensions.cs`

theMickster (changes requested):
One minor but important adjustment. Dealers choice on how to implement said change.

docs/architecture/adr/0032-break-up-core.md
justindbaur: Changed it to `AddMailer`, I'm not planning on being to strict for the `Services` suffix. Things like utilities and especially ones that don't actually add anything with the suffix service it might make more sense to drop it. But generally for whole features you're adding quite a few different things that as a collection I would refer to as services but not in this example.

theMickster: Works for me. Thanks!
Addressed in a7802eeb47b02bc62cb7e7f702ec5315a2b81f3f

theMickster (approved):

PR #833, opened 2026-07-16
docs: document populating NOT NULL columns from bulk procedures

Hinton:
## Summary

Adds guidance to the SQL code-style doc on keeping a `NOT NULL` column backwards compatible during a rolling deployment when it is populated through an `OPENJSON`/TVP bulk procedure.

- New section **Populating a `NOT NULL` column from a bulk (JSON/TVP) procedure**, nested as an `#####` sub-section under **Adding a column to a table** since it continues that section's NOT NULL / backwards-compatibility discussion.
- Explains why a scalar procedure parameter's default keeps it safe but a JSON/TVP field has no per-field default, so an old server sending a payload that omits the field yields `NULL` and fails the `NOT NULL` constraint.
- Shows the breaking pattern vs. the safe `ISNULL([Column], {default})` pattern.


docs/contributing/code-style/sql.md
```diff
@@ -766,6 +766,38 @@ variants), as this can lead to unnecessary storage overhead and performance issu
 
 :::
 
+##### Populating a `NOT NULL` column from a bulk (JSON/TVP) procedure
```
theMickster: 📝 Just double-checking that we don't want this header to show in the table of contents on the page because of `toc_max_heading_level: 4`. 

I didn't see other level-5 headings so just checking 😃

docs/contributing/code-style/sql.md
```diff
@@ -766,6 +766,38 @@ variants), as this can lead to unnecessary storage overhead and performance issu
...
+
+When a `NOT NULL` column is populated through a bulk procedure that reads from `OPENJSON` or a
```
theMickster: ⚠️ The wording about the JSON filed is correct, but I don't believe that the statements about TVPs are correct and need to be taken out of this section. [CREATE TYPE](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-type-transact-sql?view=sql-server-ver17) allows for defaults and I have encouraged their usage in the past wherever possible. We have an example of this in `src/Sql/dbo/User Defined Types/OrganizationSponsorshipType.sql` in the server repo.

Moreover, `ISNULL` in the procedure body can't make a TVP procedure rolling-deployment safe. When the type on the database gains a column, an old server's binary still builds its DataTable from the old column list (DapperHelpers.ToTvp), and the call is rejected before the procedure body runs: "Trying to pass a table-valued parameter with 1 column(s) where the corresponding user-defined table type requires 3 column(s)" — even when the missing columns are nullable or have defaults (I double-checked this against my local SQL Server). 

The [ADO.NET docs](https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/table-valued-parameters) note that the TVP structure "is automatically validated," and a type can't be altered in place — only dropped and recreated. 

I did some digging with Claude in the server repo and found an example of a work-around we shipped..  AccessSecretsManager shipped with a versioned `OrganizationUserType2` + `OrganizationUser_CreateMany2` , and why the bulk procs were later converted to OPENJSON (server #4521).

I suggest that we remove TVP from the guidance.

docs/contributing/code-style/sql.md
```diff
@@ -766,6 +766,38 @@ variants), as this can lead to unnecessary storage overhead and performance issu
...
+
+A scalar procedure parameter stays backwards compatible through its own default value
+(`@Column BIT = 0`), so an old server that calls the procedure without the new argument still works.
+A JSON or TVP field has no equivalent per-field default: during a rolling deployment an old server
+sends a payload that omits the field, `OPENJSON` (or the TVP) yields `NULL`, and the write fails
+against the `NOT NULL` constraint.
+
+This breaks during a rolling deployment:
+
+```sql
```
theMickster, line 785: ⚠️ / ♻️ Above we state "every insert and update site" but we only have examples regarding insert operations.  We also have a challenge to consider with updates.

Looking closer at `util/Migrator/DbScripts/2026-07-23_01_AddPamOrganizationUser.sql` for the new PAM feature, we have a potential problem. During the rolling-deployment window, an old server's binary does not have the AccessPam property, so its serialized JSON omits it, OPENJSON yields NULL, and ISNULL(..., 0) writes 0 over any AccessPam = 1 a new server had saved. (The EF path is incidentally safe — an old EF model just doesn't include the column in its UPDATE.) 
How likely is it that we hit this? Low but this is possible. 

Instead, I think the recommended code for columns should be: 
- Generic: 
`SET OU.[Column] = ISNULL(OUI.[Column], OU.[Column])`
- Specifically: 
```sql
[StatusNew] = OUI.[StatusNew],
[AccessPam] = ISNULL(OUI.[AccessPam], OU.[AccessPam])
```

Thoughts?

theMickster (changes requested):
I think we have some improvements to make. Happy to discuss as needed.

docs/contributing/code-style/sql.md
Hinton: I put it under that section since it's just a sub-section. We could promote it as a sibling or change toc limit.

docs/contributing/code-style/sql.md
Hinton: Fair, and seems like a good addition.

docs/contributing/code-style/sql.md
Hinton: Dropped.

docs/contributing/code-style/sql.md
theMickster: I think it's in a good spot and likely best to keep it out of the TOC for now. All good 😃

theMickster (approved):

PR #845, opened 2026-08-16
Add scripting language documentation

trmartin4:
## 📔 Objective

Adds documentation of how to approach scripting language choices to our contributing docs, in a new page under Code Style https://a9577d58.contributing-docs.pages.dev/contributing/code-style/scripting.

### DIFF
--- a/docs/contributing/code-style/scripting.md
+++ b/docs/contributing/code-style/scripting.md
@@ -6,9 +6,10 @@ Scripts committed to a Bitwarden repository must use the scripting language alre
 that repository. New scripts should not introduce a different language based on personal preference
 or on what an AI assistant happened to generate.
 
-A proliferation of scripting languages across a repository increases the support burden and widens
-the range of skills anyone maintaining it later must hold. Immediate inconvenience of writing in the
-established language is the price for long-term supportability.
+A proliferation of scripting languages within a repository and between repositories with similar
+code languages and frameworks increases the support burden and widens the range of skills anyone
+maintaining it later must hold. Immediate inconvenience of writing in the established language is
+the price for long-term supportability.
 
 ## Established languages per repository
 
--- a/docs/contributing/code-style/scripting.md
+++ b/docs/contributing/code-style/scripting.md
@@ -1,6 +1,6 @@
 # Scripting
 
-## Our stance
+## Scripting language choice
 
 Scripts committed to a Bitwarden repository must use the scripting language already established for
 that repository. New scripts should not introduce a different language based on personal preference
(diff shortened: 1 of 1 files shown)


docs/contributing/code-style/scripting.md
```diff
@@ -0,0 +1,31 @@
...
+maintaining it later must hold. Immediate inconvenience of writing in the established language is
+the price for long-term supportability.
+
+### Established languages per repository
+
+Defer to what a repository already uses. The current defaults in our core repositories are:
+
+| Repository family   | Scripting language |
+| ------------------- | ------------------ |
+| `clients`           | Node               |
```
quexten, line 20: @trmartin4 Currently, we have existing scripts that cannot be migrated because they are part of the release builds. For example: https://github.com/bitwarden/clients/blob/main/apps/desktop/resources/linux-wrapper.sh is part of the linux release builds, and the release builds do not have a node runtime to launch the wrapper (and if we would use node, it would be *significantly* worse for the runtime memory usage). 

Just so we are clear on the scope of this, this is for build / dev scripts, and does not apply to components of the released builds, should they need scripts, right?

I'm asking this because claude is bound to misinterpret these rules to always apply them during code review.

docs/contributing/code-style/scripting.md
```diff
@@ -0,0 +1,31 @@
+# Scripting
```
withinfocus, line 1: 💭 This is more of a general "contributing" page to me (one level up). There's no code style here.

docs/contributing/code-style/scripting.md
```diff
@@ -0,0 +1,31 @@
...
+Scripts committed to a Bitwarden repository must use the scripting language already established for
+that repository. New scripts should not introduce a different language based on personal preference
+or on what an AI assistant happened to generate.
+
+A proliferation of scripting languages within a repository and between repositories with similar
+code languages and frameworks increases the support burden and widens the range of skills anyone
+maintaining it later must hold. Immediate inconvenience of writing in the established language is
+the price for long-term supportability.
+
+### Established languages per repository
```
withinfocus, line 14: 🎨 I would think it's easier to maintain this in the repos and not here; this can be replaced with a note on where to look to find that decision inside one.

theMickster: I agree that this feels easier to maintain in our repos. Also, if we create a consistent section in our markdown with exceptions, then that should help facilitate consistent Claude code reviews of our scripts. Thus alleviating @quexten's concern [here](https://github.com/bitwarden/contributing-docs/pull/845/changes#r3794317672).

withinfocus:
@trmartin4 could we wrap up the minor items here and get this merged? I can do it if you'd like. I would like to refer to these docs 😜

PR #847, opened 2026-08-28
Security Principles Page Is Dead Link for Security Skill

prograhamming:
## 🎟️ Tracking

No tracking. Came up from a conversation in slack in team eng ai channel

## 📔 Objective
`bitwarden-security-context/SKILL.md` has a broken page. Fixing this to help the model have access to a security principles link within the [skill](https://github.com/bitwarden/ai-plugins/blob/main/plugins/bitwarden-security-engineer/skills/bitwarden-security-context/SKILL.md), but also give a high level summary of each of the principles. 

## 📸 Screenshots

**Before:**
image

**After:**
image

### DIFF
--- a/docs/architecture/security/principles/03-limited-security-on-semi-compromised.mdx
+++ b/docs/architecture/security/principles/03-limited-security-on-semi-compromised.mdx
@@ -1,15 +1,15 @@
 # P03 - Limited security for vaults on semi-compromised devices
 
+A semi-compromised device is one where malware exists in User Space but has not breached Kernel or
+OS-level protections. On such devices, clients must leverage available protections to prevent
+malware from accessing plaintext vault data while the vault is unlocked.
+
 :::info Note
 
 This principle applies only to unlocked vaults. Refer to [P02](./locked-vault-is-secure) for details
 on protections for locked vaults.
 
 :::
-
-A semi-compromised device is one where malware exists in User Space but has not breached Kernel or
-OS-level protections. On such devices, clients must leverage available protections to prevent
-malware from accessing plaintext vault data while the vault is unlocked.
 
 - **Technical controls** (e.g., data compartmentalization or HSMs): Clients should maximize the use
   of Kernel/OS-level protections or other available system mechanisms to safeguard vault data.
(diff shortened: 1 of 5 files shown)


withinfocus (changes requested):
The content is in the individual pages though -- we just need an index page that provides links to all sub-pages.

sognefej:
> The content is in the individual pages though -- we just need an index page that provides links to all sub-pages.

I was picturing how we do the ADRs page: https://contributing.bitwarden.com/architecture/adr/#adrs

With maybe keeping the principles section.

docs/architecture/security/principles/03-limited-security-on-semi-compromised.mdx
```diff
@@ -1,16 +1,16 @@
 # P03 - Limited security for vaults on semi-compromised devices
 
+A semi-compromised device is one where malware exists in User Space but has not breached Kernel or
```
withinfocus: ❓ Why did these move?

docs/architecture/security/principles/index.mdx
```diff
@@ -0,0 +1,21 @@
...
+data, and external readers who want to understand Bitwarden's security model.
+
+Principles are the overarching philosophies and commitments that guide Bitwarden's approach to
+security. They are not actionable rules. They state what Bitwarden aims to achieve and why, and they
+serve as the justification for the concrete [security requirements](../requirements.mdx) that
+follow. Each principle carries an identifier (P01, P02, and so on) that the rest of this
+documentation references directly, as described in
+[the security section's conventions](../index.mdx#references).
+
+Several principles define the device compromise level they assume. P02 governs locked vaults, while
```
withinfocus: ⛏️ Drop this section, in case something drifts and it becomes inaccurate given the duplication.

withinfocus (changes requested):

docs/architecture/security/principles/03-limited-security-on-semi-compromised.mdx
prograhamming: Reverted — P03 and P04 are back to their original order (byte-identical to `main`), so this PR is a single new file again.

For context on why they moved: the `DocCardList` on the new index page excerpts each card description from the page's first block, and on these two pages that is the admonition. Both cards rendered as "This principle applies only to unlocked vaults. Refer to P02 for details" rather than their own principle. That is cosmetic and not worth the churn here — if we want to fix the card text later, a `description` in the frontmatter does it without touching the prose.

docs/architecture/security/principles/index.mdx
prograhamming: Removed this section.

withinfocus (approved):

PR #858, opened 2026-09-04
feat(sso): Add IdP encryption configuration step

enmande:
## 🎟️ Tracking

[PM-43087](https://bitwarden.atlassian.net/browse/PM-43087)

## 📔 Objective

Adds documentation for local SimpleSaml assertion encryption configuration.

See also: `server` PR [8319](https://github.com/bitwarden/server/pull/8319). These should be merged simultaneously.

[PM-43087]: https://bitwarden.atlassian.net/browse/PM-43087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/getting-started/server/sso/index.md
+++ b/docs/getting-started/server/sso/index.md
@@ -23,88 +23,87 @@ we have pre-configured in an `idp` Docker container for easy setup.
 
 ### Configure the IdP
 
-1.   Open your local web client and navigate to your organization → Settings → Single Sign-On.
-2.   Tick the "Allow SSO authentication" box.
-3.   Come up with and enter an SSO Identifier.
-4.   Select "SAML 2.0" as the SSO type. Don't save or exit this page yet, you'll need to come back
-     to it later.
-5.   Open a new terminal and navigate to the `dev` folder in your server repository, e.g.
-     ```bash
-     cd ~/Projects/server/dev
-     ```
-6.   Open your `.env` file and set the following environment variables using the "SP Entity ID" and
-     "Assertion Consumer Service (ACS) URL" values from the SSO configuration page opened in step #4
-     above:
-
-     ```bash
-     IDP_SP_ENTITY_ID={SP Entity ID}
-     IDP_SP_ACS_URL={ACS URL}
-     ```
-
-     :::note
-
-     You should have created this `.env` file during your initial server setup. You can refer back
-     to the `.env.example` file if required.
-
-     :::
-
-7.   (Optional) You may generate a certificate to sign SSO requests. You can do this with a script
-     made for your OS of choice.
-
-     ```bash
-     # Mac
-     ./create_certificates_mac.sh
-
-     # Windows
-     .\create_certificates_windows.ps1
-
-     # Linux
-     ./create_certificates_linux.sh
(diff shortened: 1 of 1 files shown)


docs/getting-started/server/sso/index.md
```diff
@@ -78,13 +78,28 @@ we have pre-configured in an `idp` Docker container for easy setup.
     [here](https://github.com/kenchan0130/docker-simplesamlphp#advanced-usage) for more information
     about customizing this file.
 
-9.  Start the docker container:
+9.  Make a copy of the provided `saml20-sp-remote.php.example` file, which contains the
+    configuration for the IdP's assertion encryption capabilities.
```
eliykat, line 82: I suggest this is marked as optional, as you don't need it unless you want to test assertion encryption specifically.

EDIT: that might need the server PR to be updated so that Aspire doesn't try to bind a non-existent file though... let me know if you disagree.

enmande: @eliykat I agree. I would prefer it to be optional. I've wrestled a bit with how to handle this while accommodating both Aspire and Docker Compose, as their capabilities are somewhat different.

It should not be necessary to configure assertion behavior outside of testing these scenarios, as you point out. But we would like to offer that capability in a consistent way.

Aspire is the easier to accommodate by skipping the file bind if it does not exist, or by automating a copy if it does not exist in C#. Docker Compose is less flexible. It requires a file's unconditional existence to add it to bind volumes, else it makes a directory behind the scenes which could be problematic to debug later. For Compose binding support, the PHP file must exist, or we accept the directory-creating default behavior.

I have an idea to add an entrypoint in the `idp` image definition for Docker Compose. It would script a copy of the `*.example` to an un-configured `*.php` file if it does not exist, and then bind it to the container's `metadata` volume. It would be silent for developers, but would force the existence of the (un-tracked) `*.php` file. If configuration is desired, the un-tracked, auto-generated PHP file could then be edited in place. If we do this, I think Aspire behavior should be the same: use C# to check/copy the file for the bind volume.

What do you think about such a solution? It works nicely in local testing, but is a necessary deviation from, e.g., how we handle `authsources.php`, as it is required in all scenarios to use `idp`.

eliykat: I see, that's a bit tricky. Your suggestion is reasonable but I might be sending you down a rabbit hole to support different configuration options across our (unfortunately) multiple competing dev setup environments. It's your call, but I'm also OK with just making it mandatory so everyone has a consistent setup (and we don't have to go editing docker compose definitions here :)

eliykat (approved):

docs/getting-started/server/sso/index.md
JaredSnider-Bitwarden: I agree with making mandatory as this is is a simple additional step to add consistency + new capability we've discovered that auth needs for testing some SAML encryption scenarios locally. If this becomes a pain point, we can iterate at that time.

JaredSnider-Bitwarden (approved):

PR #859, opened 2026-09-08
docs: add ADR 0032 for consolidating clients, mobile, and SDK into a monorepo

coroiu:
## 🎟️ Tracking

No ticket. Drafted from a 12-month investigation into cross-repo breaking-change pain along `server` → `sdk-internal` → `clients`/`android`/`ios`.

## 📔 Objective

**Adds ADR 0032 (Proposed): merge `clients`, `android`, `ios`, and `sdk-internal` into one repo.**

The problem, measured over 2025-09-01 to 2026-08-31:

- **One in five `sdk-internal` merges is breaking** (149 of 754), and every one ships ungated. npm and Maven publish on every merge; the breaking-change check only comments.
- **Breaks never arrive alone.** Every breaking merge had another within 14 days. Consumers inherit bundles of unrelated breaks, not one change at a time.
- **68 forced downstream fixes** (clients 18, android 28, iOS 22), worsening as monthly SDK authors grew from ~14 to ~29.

**A monorepo kills the concurrent-bundle problem by construction:** a breaking change and its fixes land in one commit, so there is no published-then-adopted gap for breaks to pile up in.

**Why this over the cheaper options** (enforcing merge gate, additive-only, surface design, ownership): those cost less but lean on discipline that decays as more teams touch the SDK. The monorepo's guarantee is structural and holds regardless of headcount.

**What it does not fix, stated up front:** it is the highest-cost option, it does not remove the mobile adaptation work, and it does not close the run-time forward-compatibility gap for the deployed app fleet. The ADR records all three.

Status is **Proposed**: a suggestion to debate toward consensus, not a settled call.

## 📸 Screenshots

N/A (docs only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/architecture/adr/0032-monorepo-for-clients-mobile-and-sdk.md
+++ b/docs/architecture/adr/0032-monorepo-for-clients-mobile-and-sdk.md
@@ -1,0 +1,172 @@
+---
+adr: "0032"
+status: Proposed
+date: 2026-09-07
+tags: [clients, mobile, sdk]
+---
+
+# 0032 - Consolidate clients, mobile, and the SDK into a monorepo
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+## Context and problem statement
+
+The SDK (`sdk-internal`) is the shared core consumed by `clients`, `android`, and `ios`. Changes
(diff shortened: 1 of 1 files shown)


coroiu:
This was just an experiment with Claude. Removing

PR #842, opened 2026-08-04
ADR 0034 documentation standard publication

MGibson1:
## 📔 Objective

Approve and publish unified documentation standard.

Related PRs:

- [**ADR 0034 proposal**](https://github.com/bitwarden/contributing-docs/pull/841)
- **[template repository update](https://github.com/bitwarden/template/pull/204)**
- **[ai-plugins skill for document updates](https://github.com/bitwarden/ai-plugins/pull/193)**

PR #841, opened 2026-08-03
Introduce documentation standardization for at all communication levels

MGibson1:
## 📔 Objective

Investigate and recommend a direction for documentation standards at Bitwarden.

PR #862, opened 2026-09-10
Establish the principles of a service-oriented architecture

mike-gorman-bitwarden:
Establishes the principles services must adhere to in a service-oriented architecture.


docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,206 @@
+---
+adr: "0035"
+status: Proposed
+date: 2026-09-10
+tags: [server, sdk]
```
theMickster: ⛏️/♻️  The `sdk` is used to represent the Rust code in the `sdk-internal` repo which is a client sdk.

```suggestion
tags: [server, server-sdk]
```

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,206 @@
...
+The following decisions govern what a service is and how it behaves.
+
+1. Service boundaries `MUST` derive from data ownership, not from team structure.
+2. Every resource `MUST` have exactly one owning service, and that service is the only process that
+   reads or writes its data store.
+3. Services `MUST` be built on the `Bitwarden.Server.Sdk` package.
+4. Services `MUST` document their APIs in [OpenAPI format](https://www.openapis.org/) and conform to
+   **API Standards**.
+5. Services `MUST NOT` make breaking changes.
+6. Services `MUST` provide a **service client** for clients and other services (a.k.a. "consumers")
```
theMickster: ♻️ In the opening paragraph, we define consumers. Let's consider using that word instead of redefining it and adding the 'a.k.a' phrase.

```suggestion
6. Services `MUST` provide a **service client** for consumers.
```

docs/architecture/adr/0035-service-oriented-architecture.md
theMickster, line 1: ❓ We introduce the word `tenant` here but we don't currently use that word in any domain nomenclature (to my knowledge). Is that something we are prepared to introduce as part of the service oriented architecture? Feels out of place. That's all

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+The following decisions govern what a service is and how it behaves.
+
+1. Service boundaries `MUST` derive from data ownership, not from team structure.
+2. Every resource `MUST` have exactly one owning service, and that service is the only process that
+   reads or writes its data store.
+3. Services `MUST` be built on the `Bitwarden.Server.Sdk` package.
+4. Services `MUST` document their APIs in [OpenAPI format](https://www.openapis.org/) and conform to
+   **API Standards**.
+5. Services `MUST NOT` make breaking changes.  Changes that *would* be breaking should follow the
+   API versioning process as outlined by API Standards.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Clarified that changes that *would* be breaking `MUST` follow the API versioning process.  What that means in practice is that the API that *might* have been changed in some breaking way ends up *not* changed and, instead, a new version of that API is introduced.  As a result, no breaking changes were ever made.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+  matrices on customer installations.
+
+## Decision outcome
+
+Chosen option: **service-oriented architecture**, because it keeps exactly one implementation of
+each read under the ownership of the team accountable for its rules, and because putting a resource
+behind exactly one owner is what makes tenant isolation enforceable in a single place rather than as
+a convention repeated at every call site.
+
+The following decisions govern what a service is and how it behaves.
```
theMickster: ⚠️ As mentioned in Slack chat, we have decided to migrate this block into a new page under the `docs/architecture` and, instead, refactor the decision to model more of what we wrote for [0034 - Adopt engineering documentation standard](https://contributing.bitwarden.com/architecture/adr/adopt-engineering-documentation-standard#decision-outcome) where we point readers to the new standards.

Pull Requests that should help with the refactor
1. https://github.com/bitwarden/contributing-docs/pull/841
2. https://github.com/bitwarden/contributing-docs/pull/842

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+**Cons**
+
+- Introduces a synchronous dependency between services, which must be authenticated, authorized,
+  cached, and operated.
+- Independently deployable services are independently versioned services, which means compatibility
+  matrices on customer installations.
+
+## Decision outcome
+
+Chosen option: **service-oriented architecture**, because it keeps exactly one implementation of
```
theMickster: ⚠️ I think this is too excessive.  I think repeating the argument about tenant isolation detracts from the actual decision. 

Let's consider a direct, simple decision. Perhaps ->  

```suggestion
Chosen option: service-oriented architecture - one implementation of each read, owned by the team accountable for its rules.
```

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+
+The pieces needed to decompose already exist. `Bitwarden.Server.Sdk` is a shared MSBuild SDK package
+consumed by fifteen projects in `server`. Command-query separation is established at scale per
+[ADR-0008](./0008-server-CQRS-pattern.md), with several hundred single-operation command and query
+classes already in the tree, and the newest service is factored into endpoints, handlers, and
+commands. What is missing is not a mechanism but an agreement: a definition of what a service is,
+what it owns, and how a boundary is crossed. Absent that, each extraction is negotiated from scratch
+and the boundaries drift apart.
+
+Deciding nothing has a specific cost. Services will get extracted regardless, because teams need
```
theMickster: ❌ Please remove this sentence. It's a classic AI epigrams that detracts from the value of the problem statement. `Deciding nothing has a specific cost. `

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+
+- **Version skew becomes a supported condition.** Independently deployed services mean compatibility
+  matrices on customer installations, with no rollback available on a customer's own hardware.
+  Decision 5 is what keeps this tractable, and it is a permanent obligation rather than a migration
+  cost.
+- **A synchronous dependency now exists where none did.** It has to be authenticated, authorized,
+  observed, and operated, and a dependency's unavailability becomes a caller's failure mode.
+- **Service-to-service authentication has to be built for cloud.** The existing internal grant has
+  only ever been registered for self-hosted deployments, because cloud has had no service-to-service
+  path. Decomposition creates that path.
```
theMickster: ❌ Please remove this sentence. It's a classic AI epigrams that detracts from the value of the statement.
`Decomposition creates that path.`

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+     document the security ramifications of a stale copy (due to messaging lag, event processing
+     failures, etc.).
+10. Services `MUST` publish events as a matter of course using the "transactional outbox" pattern,
+    regardless of whether there are any known consumers.
+
+### Positive consequences
+
+- A resource has one owner, so a schema change is reasoned about locally instead of across the
+  entire application.
+- Decision 2 makes an owning service the only reader of its own store, which is what allows tenant
```
theMickster: ♻️ Please consider a refactor. This reads to me as an overstatement of the decision outcome in the above section.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+   - Serving results from cache `MUST NOT` bypass authorization the owning service would otherwise
+     enforce.
+8. Services that need to read, write, or validate data owned by another service `SHOULD` do so via
+   the service's published service client.
+9. A service `MAY` hold a local copy of another service's data only with a recorded justification
+   (e.g. a measured hot-path volume, a stated availability requirement, etc.).
+   - Any service holding a local copy `MUST` enforce the owner's row-level security on that copy and
+     document the security ramifications of a stale copy (due to messaging lag, event processing
+     failures, etc.).
+10. Services `MUST` publish events as a matter of course using the "transactional outbox" pattern,
```
theMickster: ♻️ Please consider refactoring out this phrase `as a matter of course` because it is textbook archaic AI filler idioms.

theMickster (changes requested):
I think we have a good start and are 80% of the way there to a great ADR.

I have noted some specific concerns with textbook AI writing fingerprints that should be addressed; but let's be sure to re-read the whole document to pull out anything else.  

We have a lot of the _[X] rather than [Y]_ writing style that we should review and consider if each one is required.  I realize that these statements are important parts of an ADR because we are comparing decision options, but there's like 10+ uses of the writing style and they cannot all be required for the ADR to stand strong. 

I also think we can cut many of the overly verbose and run-on statements throughout the ADR without loosing technical merit.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: That phrasing came directly from me and is the phrasing I have used for years whenever I am talking to people about the role of events and publishing messages in a service-oriented architecture.

theMickster: Well then Claude must have been trained on your writing style; hope you have a kickback! 😜

Perhaps `by default` instead of `as a matter of course` is a drop-in replacement that is a smoother read? 🤔

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+    regardless of whether there are any known consumers.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Deferred deliberately.  Enumerating constraints on authorized event listeners is beyond the scope of this ADR.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,207 @@
...
+Deciding nothing has a specific cost. Services will get extracted regardless, because teams need
+independent deployment, and they will be extracted with divergent answers to the same questions —
+who may read this table, what happens when the owner is unavailable, whether a copy is acceptable.
+Reconciling those answers after the fact is far more expensive than agreeing them once.
+
+## Considered options
+
+- **Status quo:** one application over one shared store, with logical separation by convention.
+- **Modular monolith:** enforce boundaries in code (module ownership, architecture tests) while
+  keeping the single store and single deployment.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: References added to the canonical page documenting SOA principles.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Unintentional and, in any event, certainly out of scope for this ADR.  Corrected.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Reorganized.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Removed.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Removed.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Refactored.  782788d679e86f6796136db2854884016325d4e8

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: I adopted "as a matter of course" years ago as shorthand to emphasize that the decision to publish events is not based on whether or not you are aware of some specific consumer or use case that wants the event, but that is stated explicitly here already ("regardless of whether there are any consumers") and the phrase did not clarify "for what".  A service might publish an event for every read under the old phrasing.

Removed "as a matter of course" and clarified "for what": `for every state change`

docs/architecture/service-oriented-architecture/services.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Services
+
+**Audience:** Bitwarden engineers and AI agents building, extracting, or reviewing a server-side
+service.
+
+This page is the living standard for how Bitwarden's services relate to one another: what a service
+owns, how a boundary is crossed, and what every service publishes. It was adopted in
+[ADR-0035](../adr/0035-service-oriented-architecture.md). The rules below evolve by pull request
+without superseding that decision.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed.  ADR will land as `Accepted` upon merging.

docs/architecture/adr/0035-service-oriented-architecture.md
theMickster: Even better than my suggestion. Thanks!

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,170 @@
...
+`SHOULD`, `SHOULD NOT`, `MAY`) deliberately. Anything marked `MUST` or `MUST NOT` is not negotiable
+at team level; a team that needs an exception brings the case to the architecture group. The terms
+_service_, _consumer_, _service client_, and _row-level security_ are defined in the
+[service-oriented architecture standard](../service-oriented-architecture/services.md).
+
+## Context and problem statement
+
+The server is one monolithic application over one monolithic database. Any code path can join
+across any domain, so a domain's data has no enforceable owner: the database is the integration
+contract, and every team is coupled to every other team's tables. Three consequences follow:
```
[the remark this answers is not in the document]
theMickster: Bit by the old prettier dragon 🐉  👎🏼

Double-check that you have prettier installed via npx and it's working properly in tandem with `.husky/pre-commit`.  Every so often this goes sideways with me too 😢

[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,170 @@
...
+**Cons**
+
+- Introduces a synchronous dependency between services, which must be authenticated, authorized,
+  cached, and operated.
+- Independently deployable services are independently versioned services, which means compatibility
+  matrices on customer installations.
+
+## Decision outcome
+
+Chosen option: **Service-Oriented Architecture**
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Dropped intentionally.  Options were summarized already; we don't need to repeat the "pros" here.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Notation
+
+This ADR uses [RFC 2119](https://www.rfc-editor.org/info/rfc2119/) keywords (`MUST`, `MUST NOT`,
+`SHOULD`, `SHOULD NOT`, `MAY`) deliberately. Anything marked `MUST` or `MUST NOT` is not negotiable
+at team level; a team that needs an exception brings the case to the architecture group.
+
+## Context and problem statement
+
+The server is one monolithic application over one monolithic database. Any code path can join across
```
theMickster, line 20: ⚠️ This sentence is difficult to understand. The repeated phrasing and punctuation overload are it's main weaknesses. 
```
Any code path can join across any domain, so a domain's data has no enforceable owner: the database is the integration contract, and every team is coupled to every other team's tables.
```

💡 Perhaps something similar to one of these alternatives

```
No boundary enforces ownership of a domain's data, so any code path can read or join across domain boundaries. The database itself is the integration contract.
```

-OR-
```
Any code path can join across domain boundaries, leaving data with no enforceable owner. The database serves as the integration contract, coupling teams directly to one another's tables.
```

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+4. Services `MUST` document their APIs in [OpenAPI format](https://www.openapis.org/) and conform to
+   **API Standards**.
+5. Services `MUST NOT` make breaking changes. Changes that _would_ be breaking `MUST` follow the API
+   versioning process as outlined by API Standards.
+6. Services `MUST` provide a **service client** for consumers.
+7. Service clients `SHOULD` make use of a network cache to mitigate performance issues.
+   - Any cache used `MUST` be owned and invalidated by the owning service.
+   - Serving results from cache `MUST NOT` bypass authorization the owning service would otherwise
+     enforce.
+8. Services that need to read, write, or validate data owned by another service `SHOULD` do so via
```
theMickster, line 122: ♻️ Rule #8 has an awkward noun reference that we need to clarify. I think we have the word "service" too many times in this rule. 

```
Services that need to read, write, or validate data owned by another service `SHOULD` do so via the service's published service client.
```

💡 Perhaps the following:
```
Services that need to read, write, or validate data owned by another service `SHOULD` do so via the owner's published service client.
```

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+   the service's published service client.
+9. A service `MAY` hold a local copy of another service's data only with a recorded justification
+   (e.g. a measured hot-path volume, a stated availability requirement, etc.).
+   - Any service holding a local copy `MUST` enforce the owner's row-level security on that copy and
```
theMickster: ♻️ I think we have a similar stacked noun problem with the word "copy" used three times. 

```
Any service holding a local copy MUST enforce the owner's row-level security on that copy and document the security ramifications of a stale copy (due to messaging lag, event processing failures, etc.).
```

💡 Perhaps the following:
```
Any service holding a local copy `MUST` enforce the owner's row-level security on that data and document the security ramifications of stale reads (due to messaging lag, event processing failures, etc.).
```

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+- Audit and future integrations read one event stream that already exists.
+- Local copies stay available where warranted, with the justification and the staleness consequences
+  recorded where the copy is introduced.
+
+### Negative consequences
+
+- **Version skew becomes a supported condition.** Independently deployed services mean compatibility
+  matrices on customer installations, with no rollback available on a customer's own hardware. Rule
+  5 keeps this tractable, and it is a permanent obligation.
+- **A synchronous dependency now exists where none did.** It has to be authenticated, authorized,
```
theMickster, line 165: ♻️ The double "and" in the following sentence squishes two ideas into one sentence. 

```
It has to be authenticated, authorized, observed, and operated, and a dependency's unavailability becomes a caller's failure mode.
```

💡 Perhaps the following:
```
It has to be authenticated, authorized, observed, and operated. A dependency's unavailability becomes a caller's failure mode.
```

theMickster (changes requested):
I pulled down the branch and re-read it after running the project. I think we have a few grammatical improvements below to make. 

One point of confusion that I am unsure whether it matters to solve in this document is that we reference 'rule # ' over 5 times in the last two sections.  Without anchors or strong context (e.g. a rule name), a reader is forced to scroll up and down the document trying to locate what rule ## is.  For an ADR, I don't think that's a must-fix. However, we need to remember to address this in the documentation changes that we are proposing in #863.  I'm not sure if that means each rule needs a formal name, we use anchors in the documents, or (maybe both) we need to make this easier to consume.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done. 717b124511e5b6bbe553e3a628f7b098df7bda16

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done. 717b124511e5b6bbe553e3a628f7b098df7bda16

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done. 717b124511e5b6bbe553e3a628f7b098df7bda16

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done. 717b124511e5b6bbe553e3a628f7b098df7bda16

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Per Slack conversation, removed everything but the ADR.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+     enforce.
+8. Services that need to read, write, or validate data owned by another service `SHOULD` do so via
+   the service's published service client.
+9. A service `MAY` hold a local copy of another service's data only with a recorded justification
+   (e.g. a measured hot-path volume, a stated availability requirement, etc.).
+   - Any service holding a local copy `MUST` enforce the owner's row-level security on that copy and
+     document the security ramifications of a stale copy (due to messaging lag, event processing
+     failures, etc.).
+10. Services `MUST` publish events for every state change using the "transactional outbox" pattern,
+    regardless of whether there are any known consumers.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Deliberate.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+   versioning process as outlined by API Standards.
+6. Services `MUST` provide a **service client** for consumers.
+7. Service clients `SHOULD` make use of a network cache to mitigate performance issues.
+   - Any cache used `MUST` be owned and invalidated by the owning service.
+   - Serving results from cache `MUST NOT` bypass authorization the owning service would otherwise
+     enforce.
+8. Services that need to read, write, or validate data owned by another service `SHOULD` do so via
+   the service's published service client.
+9. A service `MAY` hold a local copy of another service's data only with a recorded justification
+   (e.g. a measured hot-path volume, a stated availability requirement, etc.).
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: If services want to hold a local copy, they `MUST` justify it.

docs/architecture/adr/0035-service-oriented-architecture.md
theMickster: Thank you!! That reads so much better.

docs/architecture/adr/0035-service-oriented-architecture.md
theMickster: Thanks!

docs/architecture/adr/0035-service-oriented-architecture.md
theMickster: Thanks!

theMickster (dismissed):

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,165 @@
...
+  made.
+- Audit and future integrations read one event stream that already exists.
+- Local copies stay available where warranted, with the justification and the staleness consequences
+  recorded where the copy is introduced.
+
+### Negative consequences
+
+- **Version skew becomes a supported condition.** Independently deployed services mean compatibility
+  matrices on customer installations, with no rollback available on a customer's own hardware. Rule
+  5 keeps this tractable, and it is a permanent obligation.
```
mzieniukbw: ❓ Does this allow for eventual deprecation and removal of a version ?

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,168 @@
...
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+## Notation
+
+This ADR uses [RFC 2119](https://www.rfc-editor.org/info/rfc2119/) keywords (`MUST`, `MUST NOT`,
+`SHOULD`, `SHOULD NOT`, `MAY`) deliberately. Anything marked `MUST` or `MUST NOT` is not negotiable
+at team level; a team that needs an exception brings the case to the architecture group.
+
+## Context and problem statement
```
withinfocus, line 18: 🎨 Given the considered options, I think we could expand on why we see this as a problem at all. We want independent deployments, but -- perhaps briefly -- we can explain what that does for us as R&D overall.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: That's a great question.  The ADR doesn't explicitly say how new services will be versioned going forward but I will strongly advocate services living in repos of their own, with semantic version numbers evolving independently, as well, as new features are added and bugs are fixed.  However, [upcoming standards for internal APIs](https://github.com/bitwarden/contributing-docs/pull/865) mandates services never make breaking changes which means, "the latest version of everything" is supposed to always work together; that is our obligation.  (This same ADR formalizes how APIs _do_ get formally deprecated and eventually removed.)

But, as far as versions of the _services_ goes, a self-host customer initial install is "everything" and, to get any bug fix, the process is also the same: upgrade to the latest version of everything.  In this view, every bug is forward-fix only; there are no "major versions" or release branches to maintain: just a never-ending stream of enhancements and bug fixes.

Now to your question: if you are talking about deprecating and, eventually, removing some version of some service, I would say we _can_ do that but I am not sure it is necessary.  We just continuously add features and fix bugs and, each time we do, we are publishing new versions of the services and customers can just sit on whatever version they are on or they can upgrade to the latest (of everything) to get all of the latest bug fixes across all services.  From this point of view, every version of every service that is not the latest is deprecated (because it contains known bugs that have since been fixed).

Having said that, this is my "I'm new here" view of how this should work without the first conversation with Product Management so they may take a different view of how versioning this federation of services should work and may have reasons to retain the current `YYYY.XX.ZZ` versioning scheme.

Does that help at all?

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Service-oriented architecture
+
+**Pros**
+
+- One implementation of each read, owned by the team that owns the rules it enforces.
+- Organization scoping is enforced once, by the owner.
+- Local copies remain available where they are genuinely warranted, as a recorded exception.
+
+**Cons**
```
withinfocus, line 92: 🎨 One more: latency, and N+ complexity. This can have a material impact on SLOs.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,168 @@
...
+
+- Introduces a synchronous dependency between services, which must be authenticated, authorized,
+  cached, and operated.
+- Independently deployable services are independently versioned services, which the release pipeline
+  has to keep shipping as one coordinated set.
+
+## Decision outcome
+
+Chosen option: **Service-Oriented Architecture**
```
withinfocus:
```suggestion
Chosen option: **Service-oriented architecture**.
```
⛏️ per our norms.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,168 @@
...
+  observed, and operated. A dependency's unavailability becomes a caller's failure mode.
+- **Service-to-service authentication has to be built for cloud.** The existing internal grant has
+  only ever been registered for self-hosted deployments.
+- **Every extracted service is another process on the smallest tier.** Bitwarden Lite already runs
+  nine processes on one box against an operator-supplied database, so service count is priced there
+  first.
+- **Row-level security in the data layer is not yet portable.** The current implementation composes
+  T-SQL and has no Entity Framework path, so the enforcement this ADR relies on is available on SQL
+  Server only until that gap is closed.
+- **Caching is not uniformly available.** Neither full self-host nor Bitwarden Lite ships a shared
```
withinfocus: ⚠️ Depending on what "shared cache" means here, we do have customers that configure the Redis or Cosmos options exposed via configuration.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,168 @@
...
+  cache today, so rule 7's cache is cloud-only until that is addressed.
+
+### Plan
+
+- Publish **API Standards**. Rules 4 and 5 reference it normatively and it does not exist yet on
```
withinfocus, line 182: ℹ️ This will be edited / amended once in place, or its development will be clearly indicated.

withinfocus (commented):
Few things to improve upon, potentially before merging, or they can come with the later (expected) ratification.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done. 48be7c57a42bd3cee15738a2f582b42d6023beea

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Done. 48be7c57a42bd3cee15738a2f582b42d6023beea

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Upated. 48be7c57a42bd3cee15738a2f582b42d6023beea

withinfocus (dismissed):
This works for me. Great beginnings!

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,180 @@
...
+4. Services `MUST` document their APIs in [OpenAPI format](https://www.openapis.org/) and conform to
+   **API Standards**.
+5. Services `MUST NOT` make breaking changes. Changes that _would_ be breaking `MUST` follow the API
+   versioning process as outlined by API Standards.
+6. Services `MUST` provide a **service client** for consumers.
+7. Service clients `SHOULD` make use of a network cache to mitigate performance issues.
+   - Any cache used `MUST` be owned and invalidated by the owning service.
+   - Serving results from cache `MUST NOT` bypass authorization the owning service would otherwise
+     enforce.
+8. Services that need to read, write, or validate data owned by another service `SHOULD` do so via
```
MGibson1, line 122: **issue**: This feels like a must to me. I don't want to get in a world where we have the option to directly hit APIs to bypass guards built into a client or scatter implementation details to far-flung API call sites.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,180 @@
...
+   the owner's published service client.
+9. A service `MAY` hold a local copy of another service's data only with a recorded justification
+   (e.g. a measured hot-path volume, a stated availability requirement, etc.).
+   - Any service holding a local copy `MUST` enforce the owner's row-level security on that data and
+     document the security ramifications of stale reads (due to messaging lag, event processing
+     failures, etc.).
+10. Services `MUST` publish events for every state change using the "transactional outbox" pattern,
+    regardless of whether there are any known consumers.
```
MGibson1, line 130: Is there value in limiting this scope initially? I may be missing things, but my read of this architecture is that only deletes are needed to propagate cascades.

Additionally, this rule binds us to broad-scoped state change statements, rather than tightly typed and defined messages. The latter requires knowledge of consumer needs, but the former creates significant challenges in potentially needing to combine data streams.

MGibson1 (changes requested):

docs/architecture/adr/0035-service-oriented-architecture.md
withinfocus: Mike and I actually talked about this one and softened the language so as to not rule this out.

MGibson1: Is there something you foresee that being beneficial for? I think it's likely to produce unspoken contracts and action at a distance effects

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: The idea here is that if every service broadcasts every state change "as a matter of course", when there is a need to build this integration or that, the team that needs to do so has everything they need.  There is never a situation where, in order to make something work, team A has to talk to team B to get them to start doing something.  It is a pattern I found to be a very effective way of working.

Having said that, I think that we actually _do_ have a consumer for all of these events - and that is our audit log (a.k.a. Events Service) which, of course, wants a permanent record of everything that's happened.

In addition to that, are the consumers you mentioned - those that want to make sure they delete related data when parent resources are deleted.

So the requirement is for those three reasons:

1. To facilitate the audit log.
2. To facilitate cascade deletes.
3. To facilitate the next thing we haven't thought of so that, when it _does_ come up, it doesn't require the "upstream" team to make changes to their code.

docs/architecture/adr/0035-service-oriented-architecture.md
quexten: Thanks mike! So to clarify, you are saying that within a version there are no breaking changes to be made, but old versions can be deprecated and removed, so they do not have to be indefinitely supported?

I've added comments on the PR you link, but one of the interesting questions here is: What about security issues? Do they classify under the same category as bugs (for which breaking changes are allowed as per your linked PR)?

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,180 @@
...
+   - Serving results from cache `MUST NOT` bypass authorization the owning service would otherwise
+     enforce.
+8. Services that need to read, write, or validate data owned by another service `SHOULD` do so via
+   the owner's published service client.
+9. A service `MAY` hold a local copy of another service's data only with a recorded justification
+   (e.g. a measured hot-path volume, a stated availability requirement, etc.).
+   - Any service holding a local copy `MUST` enforce the owner's row-level security on that data and
+     document the security ramifications of stale reads (due to messaging lag, event processing
+     failures, etc.).
+10. Services `MUST` publish events for every state change using the "transactional outbox" pattern,
```
quexten, line 129: What if the data here is data that the owning team does not want to offer as a public, stable contract? I can imagine quite a few values for which this is the case. As an example, I would not want other teams to be able to consume "masterpasswordunlockdata". This is simply not safe to offer.

In client-side APIs we've seen cryptographic APIs very much misused, leading to insecure designs, even when properly documented, or even when marked as "DEPRECATED AND DANGEROUS, DO NOT USE".

quexten (changes requested):

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: It's a good point.  The standard requires events to be published (to let others know something happened) but the exact details of what is included in the event is up to the service to decide.  Indeed, we have not yet begun to establish standards for what these events should look like.  Surely, they should all share some common shape but what gets populated is up to the service.

But you _are_ right to call it a "public, stable contract".  Not public in the sense of available externally but public in the sense that it is going to be on an event bus of some sort that other services can tap into.  As such, the shape of these events needs to be something consumers can rely upon so, in that sense, it very much is a contract.

But, since this ADR is only focused on service principles, I added a blurb here to make it clear that standards for these events will be the subject of a forthcoming ADR and out of scope for this ADR.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Yeah - it's definitely a different way of thinking - because even sentences like this cease to have any meaning in a world of continuous deployment:

> ... so they do not have to be indefinitely supported?

This sentence has in mind something like release branches where a bug that is found has to be fixed in N number of releases.  But if services are just continuously receiving updates, and every one results in a new version of that service, with a new semantic version number, then there is no concept of "supporting" anything other than the current version.  If a bug is found in the current version or any previous version, it is just fixed on the current version - and a new version produced.

It is our job to fix any bug that is found.  It is the self-host customer's job to upgrade to the latest to get whatever bugs we have fixed.

> Yeah, but customers want to be able to upgrade on their own timetable.  And they do not like to get new features and bug fixes at the same time.  They want to _just_ get bug fixes.

Well I guess that is a Product Management decision but, in my view, customers do not get to dictate how they get bug fixes.  As long as the process to get them requires nothing more than a "update everything" script, I think that is not unreasonable.

The only scenario I can think of in a world of continuous deployment where there is a notion of "supporting" older versions of the software is if we came out with "Bitwarden - The Next Generation" which was like a completely new implementation based on a completely new code base.  In this case, it would be our responsibility to fix any bug found in both code bases.

> What about security issues? Do they classify under the same category as bugs (for which breaking changes are allowed as per your linked PR)?

Right.  I updated the docs to clarify that any bug - security or otherwise - `SHOULD` be fixed "in place" - even if doing so would technically be considered a breaking change.

quexten: I think the middle quote is from some other thread. I'm not sure where it came from?

Thanks for the response though!

docs/architecture/adr/0035-service-oriented-architecture.md
quexten: So it would be fine to publish something like: "The user's keys changed" (without the contents of the key change)? In that case I'm OK with that, as long as the creators / owners of the events get to define whether the changed data is contained, or some other data, or an empty marker message.

docs/architecture/adr/0035-service-oriented-architecture.md
MGibson1: Event logs are very far from an audit trail of everything that happens in the Bitwarden database. What's more is there's a layer of authorization that needs to occur on them and it's not obvious to me where that belongs.

I also want to point out that 
> To facilitate the next thing we haven't thought of so that, when it does come up, it doesn't require the "upstream" team to make changes to their code.

and 

> but the exact details of what is included in the event is up to the service to decide
are statements in conflict. You're trying to preempt what events will be needed without knowledge of what that may be.

My push back here is that this rule will result in generic, all-encompassing change events that are difficult to consume, annoying to maintain, and mostly ignored.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: > Yeah, but customers want to be able to upgrade on their own timetable. And they do not like to get new features and bug fixes at the same time. They want to just get bug fixes.

Ha!  That was me trying to anticipate the objection. ;-)

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Probably best discussed in person but I am curious to know what makes you think they would be difficult to consume?  Presumably, all events share some common top-level "shape" like `category`, `occurredAt`, `description`, and `severity` which would go a long way towards making them pretty straight-forward to consume.

As for maintenance, I would expect that most transactions go through some common framework that, in addition to persisting the changes, publishes the corresponding event in which case these events all originate from one place, which shouldn't be too hard to maintain.

And if every change is something that we need to record for auditing, then, _none_ of these events are ignored.

As for authorization, I agree that's something that needs to be worked out.

docs/architecture/adr/0035-service-oriented-architecture.md
mike-gorman-bitwarden: Definitely.  We really have not devoted too much thought to what the standards are for events that are published but, undoubtedly, it will be some JSON with a standard set of top-level fields and then a final `data` or `attributes` field that is just a dictionary of relevant fields & values with no requirement that it include the kitchen sink - especially if the data is sensitive or encrypted.

docs/architecture/adr/0035-service-oriented-architecture.md
quexten: Oh, I can see why one would have that objection, but I don't hold that view at all. 

It just happens to be that my team has to make a fair amount of changes to the cryptography layer, so I don't view the API as stable enough to be declared stable / am concerned what impact it will have / what mechanisms we will have to use to help with that. That's all.

mike-gorman-bitwarden: Well, if an API if a team is still actively iterating on an API (i.e. still under active development), I don't think "no breaking changes" rule applies to such an API.  But, in either event, this will be more relevant to the upcoming [Internal API Standards](https://github.com/bitwarden/contributing-docs/pull/865).

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,194 @@
...
+
+- Publish **API Standards**. Rules 4 and 5 reference it normatively and it does not exist yet on
+  this site.
+- Publish a client strategy page covering how a service client is generated, wrapped, versioned, and
+  cached.
+- Publish a service identity and context page covering service-to-service authentication,
+  authorization by scope, and context propagation, and build the cloud path it describes.
+- Provide an Entity Framework path for organization scoping, so rule 9 holds on all supported
+  database providers.
+- Provide a transactional outbox and a broker-free event transport, so rules 10 through 12 hold on
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed.

theMickster:
Howdy @quexten & @MGibson1 👋🏼
Do you have any remaining major concerns that blocks this from being approved in a Proposed state? No worries if so, but we have presented it to the arch council and I think we need to aim to merge it shortly (like by tomorrow if possible). 

To clarify, I am not suggesting rushing towards an Approved state of the ADR. I understand it's a meaty one. Thank you both!

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,189 @@
...
+     document the security ramifications of stale reads (due to messaging lag, event processing
+     failures, etc.).
+10. Services `SHOULD` publish events for all relevant state changes using the "transactional outbox"
+    pattern, regardless of whether there are any known consumers. Exactly what makes an event
+    "relevant", the shape of such events, the authorization model, dead-letter policies, and how
+    such events are delivered and consumed will be the subject of a forthcoming ADR and is out of
+    scope here.
+11. A service that owns resources whose lifetime depends on a resource owned by another service
+    `MUST` consume that owner's "resource deleted" events and cascade the deletion to the resources
+    it owns. An owning service is not responsible for deleting data it does not own.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed in 301a761 — rule 10 now carries a sub-bullet pinning the producer side for the delete case only:

> - A service that owns a resource other services may depend on `MUST` publish a "resource deleted" event for it, so that rule 11 is satisfiable.

That keeps the narrowing @MGibson1 asked for — deletes are what cascades actually need — and leaves the rest of rule 10 at `SHOULD` with "relevant" still deferred to the events ADR.

Not taking the second half: the Positive consequences bullets stay as written. That section describes what the architecture is for, not the keyword strength of each rule, and rewriting it to hedge against rule 10's `SHOULD` makes it read as less decided than it is.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,194 @@
...
+  recorded where the copy is introduced.
+
+### Negative consequences
+
+- **Versioning.** Independently deployable services will likely be versioned independently as well,
+  and enhancements and bug fixes will land service by service, each producing a new version of that
+  service. Rule 5 ensures a new version never introduces a breaking change, so a customer may at any
+  time run an "upgrade everything" script and take the latest of every service — which is exactly
+  what self-host installs. Our obligation is to just make sure "the latest version of everything"
+  always works.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed in 301a761 by rewording rather than restoring `MUST NOT` — `a7de27c` softened rule 5 deliberately, so the bullet was what needed to move, not the rule.

It now reads that rule 5 "makes a breaking change the exception rather than the norm," and names what absorbs the exception: where a break is genuinely unavoidable, the versioning process is what keeps an "upgrade everything" run working. The self-host point and the closing obligation are unchanged.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,194 @@
...
+2. Every resource `MUST` have exactly one owning service, and that service is the only process that
+   reads or writes its data store.
+3. Services `MUST` be built on the `Bitwarden.Server.Sdk` package.
+4. Services `MUST` document their APIs in [OpenAPI format](https://www.openapis.org/) and, if the
+   service is an internal service, such APIs `MUST` conform to the forthcoming **Internal API
+   Standards**.
+5. Services `SHOULD NOT` make breaking changes.
+   - If the changes that need to be made _would_ be breaking and the service is an internal service,
+     such services `SHOULD` follow the API versioning process as outlined by the forthcoming
+     Internal API Standards.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Good catch, and fixed in 301a761 — though not by defining "internal service", because that term can't be made to work here.

Both candidate words already name the *client-facing* API, which is the surface this ADR wants out of scope:

- **"internal"** — the server repo's own split. `ApiExplorerGroupConvention` groups operations into `public` and `internal`, served at `specs/internal/swagger.json` and uploaded by CI as `internal.json`. That internal document is the 600-odd operations our clients call over the internet.
- **"private"** — [`docs/getting-started/server/public-api.md`](https://github.com/bitwarden/contributing-docs/blob/main/docs/getting-started/server/public-api.md) on `main`, already published: "Most developers will be more familiar with the private API used by our client applications," with a table contrasting Private API against Public API.

So there are three layers — published/supported (the Public API), internet-reachable but undocumented (what the docs call the private API and the server repo calls internal), and service-to-service — and only the third is in scope for these rules.

Rules 4 and 5 now gate on **"service-to-service API"**, which this ADR already used three times for exactly this relationship, and which ADR-0036 defines as equivalent to its "Internal API". The gate also moved from the service to the API surface, since an extracted service can reasonably serve both a private API to clients and a service-to-service API to other services.

Leaving line 167's "internal grant" alone — that's the OAuth grant's actual name.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,198 @@
...
+  Server only until that gap is closed.
+- **A cache cannot be assumed to exist.** Redis and Cosmos are both exposed through configuration
+  and some self-host operators do configure one, but neither is provisioned by default and Bitwarden
+  Lite ships no cache service at all. Rule 7's caching is therefore an optimization a service may
+  find available, never a mechanism it can depend on being there.
+
+### Plan
+
+- Publish **API Standards**. Rules 4 and 5 reference it normatively and it does not exist yet on
+  this site.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Already addressed in eb7b990, though not with this wording. The Plan entry and rules 4 and 5 now all read **Service-to-Service API Standards** — "internal" already names the client-facing API in the server repo (`specs/internal/swagger.json`) and in `public-api.md` on this site, so the page in #865 has been renamed rather than the Plan entry adopting the old name.

docs/architecture/adr/0035-service-oriented-architecture.md
```diff
@@ -0,0 +1,198 @@
...
+   - Any service holding a local copy `MUST` enforce the owner's row-level security on that data and
+     document the security ramifications of stale reads (due to messaging lag, event processing
+     failures, etc.).
+10. Services `SHOULD` publish events for all relevant state changes using the "transactional outbox"
+    pattern, regardless of whether there are any known consumers. Exactly what makes an event
+    "relevant", the shape of such events, the authorization model, dead-letter policies, and how
+    such events are delivered and consumed will be the subject of a forthcoming ADR and is out of
+    scope here.
+    - A service that owns a resource other services may depend on `MUST` publish a "resource
+      deleted" event for it, so that rule 11 is satisfiable.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Right, and the qualifier was mine — fixed in 6fb9dae by dropping it:

> - A service `MUST` publish a "resource deleted" event for each resource it owns, so that rule 11 is satisfiable.

The universal reading was the intent. Rule 10 already publishes "regardless of whether there are any known consumers", so this follows the rule it hangs under rather than asking the owner to assess who depends on it — which is what line 154 says owners do not need to know.

withinfocus (approved):

MGibson1 (approved):

PR #865, opened 2026-09-15
Establish internal API standards

mike-gorman-bitwarden:
Establish the internal API standards of a service-oriented architecture.


docs/architecture/adr/0036-internal-api-standards.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Context and problem statement
+
+Service-to-service APIs are new. As server functionality is decomposed into independently deployable
+services, a service calling another service needs a wire contract, and no documented standard
+exists.
+
+One important distinction to keep in mind as we consider various options is that the APIs we are
+talking about are _internal_ APIs. We control both the service that declares them and every client
+that consumes them. While we _do_ still want to adhere to the principle of "no breaking changes", we
```
quexten, line 31: What happens if there is a vulnerability we need to fix and cannot fix without making a breaking change?

What happens if there is a cryptographic stability issue that we cannot fix without making a breaking change? (See recent key id tracking of vault items)?

docs/architecture/adr/0036-internal-api-standards.md
```diff
@@ -0,0 +1,181 @@
...
+The specific standards, including where we deviate from JSON:API, are published here:
+[Internal API standards](../server/internal-api-standards.md). That page is the living reference:
+its rules evolve by pull request without superseding this decision, and this ADR is superseded only
+if the model itself changes.
+
+### Positive consequences
+
+- One document shape across every internal API, so a consumer calling five services learns one error
+  format, one pagination scheme, and one filter grammar.
+- Versioning gives a contract a way to change shape. A breaking change becomes a new version with a
```
quexten, line 151: We have a few data points which essentially are just string encloded blobs. Examples are: vault item data bobs, encryption keys, etc. If these are changed, the json shape stays the same, but the allowed set of values in the string field changes. Would this need a newly versioned API?

docs/architecture/adr/0036-internal-api-standards.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Decision outcome
+
+Chosen option: **Adopt JSON:API selectively with versioning**.
+
+- We believe an established standard, with thoughtful answers to every API question, will be more
+  robust than any standard we might invent ourselves. It is widely adopted among some of the largest
+  SaaS vendors in the industry including [ART19](https://marketplace.apilayer.com/art19_content-api)
+  (an Amazon company) and [Datadog](https://docs.datadoghq.com/api/latest).
+- We feel strongly that internal APIs should be formally versioned. Without formal versioning, every
```
quexten: This has historically not been true, and has been solved by the +-3 versions support guarantee. If a client / server i outside of that range, we do not support that. So after three versions you are able to remove deprecated fields.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+duplicated in the URL path.
+
+## Breaking changes and versioning
+
+APIs `MUST NOT` make **breaking changes**. A breaking change is defined as follows:
+
+1. Removing (or relocating) a field.
+1. Changing the datatype of a field.
+1. Making an optional field required.
+1. Adding additional constraints to a field.
```
quexten, line 126: We will have lots of breaking changes from upcoming stability enhancements to our vault items alone. "Additional constraints on a field" each vault item will gain metadata validation that enforces that the correct key was used to encrypt the vault item. Would each of these require a new API version? 

If so we will see a burst of changes and our current API under this lens is not yet stable enough to commit to a stable contract.

quexten: Or, I guess we could define this as "intended but not enforced behavior" (a bug) and then by the next paragraph it would be enforced later?

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+> adding one.
+
+## Unrecognized fields, query parameters, and headers
+
+Sometimes a new, optional field, query parameter, or header is added to an API. Consumers upgrade to
+new clients and start passing the new fields/parameters/headers and everything is great until a
+security issue is discovered. The service is rolled back to the previous version but consumers are
+still using newer versions of the client.
+
+To keep from having to _also_ revert all of the consumers, APIs `MUST` ignore unrecognized fields,
```
quexten, line 154: We must make sure to not have requests that can be misinterpreted in this case. For example, suppose you have a request such as:

{
   key_1: abcabcabc
   key_2 (new): defdefdef
   data: [encrypt(abcabcabc,hello), encrypt(defdefdef, hello)]
}

then ignoring "key_2" will now mean that the data has permanently become inaccessible. Or, I guess we need to formalize and ban this change type?

quexten (changes requested):

docs/architecture/server/internal-api-standards.md
mike-gorman-bitwarden: Those are good points and I have done two things to try to accommodate them:

1. Soften the standard to `SHOULD NOT`.
2. Carve out an explicit exception for "security and cryptographic issues".

Having said that, when these scenarios come up, I try to keep in mind the target architecture is a fleet of independently-deployable services running in production with zero downtime (achieved via rolling updates or "rollouts").  If I have changes to make that are breaking changes, then I am faced with one of two options:

1. Roll out the breaking changes and accept that I am about to break some part of the system for some number of users for some period of time (until I can get corresponding changes rolled out to all clients/other services).
2. Try to coordinate the rollout so that the breaking changes and the corresponding client/consuming-service changes are all rolled out together.

Neither of these are good options.

- Obviously, I do not want to break the system.  And, if I have no control over forcing all customers to upgrade to new clients (e.g. CLI, mobile, etc.) then I have no control over how long the window is where the system is broken.
- And if I get into a habit of coordinating rollouts of servers, then I am losing perhaps the single-most important characteristic of my service-oriented architecture: independently-deployable services.  It is not independently-deployable if the rollout has to be coordinated with some other service or client update.

> We will have lots of breaking changes from upcoming stability enhancements to our vault items alone. "Additional constraints on a field" each vault item will gain metadata validation that enforces that the correct key was used to encrypt the vault item

This may or may not amount to a breaking change.  For example, sometimes a company will say "Starting on November 1, we will now start requiring all login attempts to required MFA."  That is a change in behavior but it is not a breaking change.  Users may now have to jump through a new hoop, but the system isn't broken.  If this is the type of change you are thinking of ("each vault item will gain metadata validation"), where the new validation is handled by the client or the UI and the user is informed that the key they are using is not correct, etc., that is not a breaking change.

> Or, I guess we could define this as "intended but not enforced behavior" (a bug) and then by the next paragraph it would be enforced later?

Right, it would be a bug, in which case it should be fixed "in place", even if doing so would technically be a breaking change.  I've updated the language accordingly in `4acc040`:

> Bugs, however, `SHOULD` be fixed "in place", without creating new versions of the API, even if the changes would technically be considered breaking changes.

No need to single out specific classes of bug — any bug qualifies.

docs/architecture/adr/0036-internal-api-standards.md
mike-gorman-bitwarden: Two places in the standard cover this, and neither requires a new version.

For a vulnerability, that's a bug, and
[Breaking changes and versioning](https://docs-adr-0036-internal-api-s.contributing-docs.pages.dev/architecture/server/internal-api-standards/#breaking-changes-and-versioning)
says:

> Bugs, however, `SHOULD` be fixed "in place", without creating new versions of the API, even if the
> changes would technically be considered breaking changes.

That's deliberately `SHOULD` and not `MAY`: versioning a security fix would leave the old, vulnerable
version in service, which is the opposite of what you want.

For something like the key ID tracking work, it depends on whether you'd call it a defect.  If it is,
the same sentence applies.  If it's better described as hardening than as fixing something broken,
the paragraph just above it covers you:

> A team `MAY` make a breaking change in place when it can account for every caller — either because
> the change has been coordinated with them, or because the rejection surfaces somewhere the caller or
> the user can act on it.

Either way: fix it in place, on the existing version.

docs/architecture/adr/0036-internal-api-standards.md
mike-gorman-bitwarden: You're right about the history — ±3 client support _is_ how we have solved this, and it does let you
drop a deprecated field once the window passes. Let me be more precise about why I don't think it
carries forward.

±3 is a guarantee about _clients_ against a platform versioned as a whole. It works because "three
versions back" is a single, fleet-wide coordinate that means the same thing everywhere. That works in
a world of regular, formal "releases" that come out every quarter, or maybe every month. But in a SOA
of independently-deployable services, each service is versioned and deployed independently. Some
slowly-changing services may run version `1.0.0` for months (or even years) while some other service
is getting bug fixes and enhancements multiple times per week — there is no shared clock against
which "three versions" means the same thing for a service that ships weekly and one that hasn't
shipped since March.

In this world, service version numbers are strictly [semantic version numbers](https://semver.org/)
where the patch number increments if the new release contains _only_ bug fixes, the minor number
increments if the new release contains one or more enhancements, and the major number increments if
there are breaking changes.

Ideally, there won't be any major version numbers as we strive to avoid making breaking changes or,
if there are, we will strive to limit them.  So counting the last N version numbers to determine when
it is okay to start dropping fields (or making them required) doesn't really work in this world.

Having said that, none of this precludes a team from enhancing APIs, rolling out updates to all
consumers, and then making the changes to the API that would, otherwise, be breaking.  Example:

1. API adds field `foo` but makes it optional.
1. Every consumer is updated to start populating `foo` and to _stop_ populating `bar`.
1. API is updated to make `foo` required and gets rid of field `bar`.

By the letter of the law, step 3 would be a "breaking change" but, in this case, the team has made
sure all consumers are populating `foo` and no longer populating `bar`.

I guess the language could be tightened up to say you can't make a field required unless you know
that all consumers are populating it but I lean toward just keeping it simple with "APIs SHOULD NOT
make breaking changes" and not try to spell out every situation where it is warranted.  The
now-softened "should not" provides the general guidance and the rest is up to the team.

docs/architecture/adr/0036-internal-api-standards.md
mike-gorman-bitwarden: Yes - by the letter of the definition, narrowing the allowed values of a string field is item 4,
"adding additional constraints to a field," so it does count as a breaking change. But to answer your
question directly: no, it would not require a newly versioned API. I've updated the rule to make that clear.  The definition is now "a breaking change is any change that causes a request that is valid today to be rejected tomorrow," followed by the same four-item list, and then:

> The verb is `SHOULD NOT` rather than `MUST NOT` because these are internal APIs and we own every caller. A team `MAY` make a breaking change in place when it can account for every caller — either because the change has been coordinated with them, or because the rejection surfaces somewhere the caller or the user can act on it.

What we want to avoid is a situation where an API is changed, requests start getting rejected, and the user has no way to fix it (i.e. to get the client or UI they are using to "do the right thing").  But
if I click save one day and get an error telling me something isn't valid and needs correcting, and
the client/UI lets me fix it, that is perfectly fine.

So for the blob cases specifically: tightening what a vault item data blob or an encryption key field
will accept is a breaking change a team may make in place, provided the rejection surfaces somewhere the user or the calling service can act on it.  If it would instead surface as a silent sync failure with no remediation path, that is the case the standard is trying to prevent — and then it does belong behind a new version.

docs/architecture/server/internal-api-standards.md
mike-gorman-bitwarden: These do not look like ordinary fields defined by a fixed schema.  This looks like an API who, by design, is expected to loop through every key present and process it which is a different sort of API than the "usual" breed I was trying to describe.  In your case, if the API failed to loop through all of the keys present, that would just be a bug.

But if this really is a "normal" API, then the scenario you are describing is this:

1. API adds a new optional field `key_2`.
1. Consumers get updated to start populating `key_2` (along with the encrypted value).
1. Service is rolled back to a previous version that does not know anything about `key_2`.
1. Consumers are _not_ rolled back; they continue sending `key_2` which the service is now ignoring.

So, yes, that is exactly what would happen in that scenario — and I want to be more precise about how bad it is than I was.  It isn't that `key_2` fails to save.  The service stores `data`, which contains a value encrypted under `key_2`, and discards `key_2` itself.  That element is then permanently undecryptable, which is materially worse than a field not persisting.

For an ordinary field I would still take availability over completeness: a system where one new field doesn't persist beats a system where save fails outright.  But that trade only holds while the damage is recoverable.  Here it isn't, and a rejected save is loud and fixable where silent unrecoverable data is neither.

On your second suggestion, I agree with the constraint.  The one thing it can't be is a runtime rule: a rolled-back service doesn't know `key_2` exists, and can't know that `data` references it, so it has no way to detect the hazard and reject the request.  It has to hold at design time instead — a new optional field has to be safe to ignore, and if dropping it would orphan data the request does store, it belongs behind a new version rather than as an optional field on the existing one.

I'm going to leave that out of the standard for now rather than spell out a case this specific, on a "less is more until it's clear more is needed" basis.  But I think it is the right rule, and if we see it come up more than once we should write it down.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+
+### Soft deletes
+
+Services `MAY` support "soft deletes" for any number of reasons. Services that do, `SHOULD` follow
+the following guidelines:
+
+- Delete APIs perform a soft delete by default and read-many APIs exclude soft-deletes by default.
+- If callers can request that a delete be "hard", callers pass `permanent=true` query parameter.
+- If callers can request that soft-deletes be included in responses, callers pass
+  `includeDeleted=true` query parameter.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Declining this one. A service that doesn't implement soft deletes hard-deletes by default, so
`permanent=true` is redundant rather than misleading — the `204` is accurate, the data really was
erased. Likewise `includeDeleted=true` against such a service: there are no soft-deleted resources to
include, so returning only live ones is the correct answer, not a silent omission.

The scenario in the finding requires a service that soft-deletes but doesn't recognize `permanent`,
and any service doing soft deletes per this section recognizes both parameters.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+  with", and "contains".
+- APIs that support filtering by an attribute of an attribute `MAY` use dot-notation (e.g.
+  `address.city`).
+
+### Multi-value parameters
+
+APIs that support specifying multiple values for a query parameter `MUST` do so via a
+comma-delimited list. This allows APIs that started with single-value query parameters to evolve to
+supporting multiple-values without changing the external-facing contract. When multiple values are
+provided, "or" query semantics `MUST` be applied.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed in `1ff5902`, scoped to where the ambiguity actually arises:

> A comma that is part of a value `MUST` be escaped with a backslash — `filter[displayName]=Smith\, Jr`
> is the single value `Smith, Jr`. This applies only where the caller is supplying multiple values or a
> [range](#ranges); a parameter that takes a single value is never split, so a comma in its value needs
> no escaping. An asterisk that is part of a value `MUST` likewise be escaped wherever the API supports
> wildcard matching.

Backslash rather than percent-encoding, because the framework decodes the query string before the
value reaches the parser — `%2C` is indistinguishable from a literal comma by then, so it can't carry
grammar. A single-value parameter is never split at all, which also disposes of the `[draft]` case: a
value that isn't parsed for delimiters is taken literally.

One correction on the details: the whitespace-stripping rule you cite applies to strings in request
bodies, not to query parameters, so it wasn't a factor here. The comma ambiguity was real on its own.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+| `Link`        | Points at the replacement.                                                                                                                    |
+
+## Divergences from JSON:API
+
+- Request and responses use the `application/json` media type, not `application/vnd.api+json`.
+- We do not support [compound documents](https://jsonapi.org/format/#document-compound-documents) or
+  the related `include` query parameter.
+- Paged responses are not required to include `links`.
+- For sparse fieldsets, callers specify the fields to include via a `fields` query parameter. We do
+  not support the `fields[{type}]` form.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Both confirmed and both added to the list in `79c6008`:

> - Unrecognized query parameters are ignored rather than rejected with `400`, and the result-shaping
>   subset is rejected with `422`.
> - Implementation-specific query parameters are not required to contain a non-`a-z` character.

On the second one, we're declining the rule rather than renaming `permanent`. Its stated rationale is
that JSON:API is "reserving the ability to standardize additional query parameters later without
conflicting with existing implementations" — which amounts to claiming every all-lowercase word in
existence. We pin the version we target, so the collision it guards against isn't one we're exposed
to.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+| ---------------- | ---------------------------------------- | ----------------------------------------------------------- |
+| `202 Accepted`   | The resource is scheduled to be deleted. | The [job](#jobs) that was scheduled to delete the resource. |
+| `204 No Content` | The resource was deleted.                | Nothing.                                                    |
+
+### Soft deletes
+
+Services `MAY` support "soft deletes" for any number of reasons. Services that do, `SHOULD` follow
+the following guidelines:
+
+- Delete APIs perform a soft delete by default and read-many APIs exclude soft-deletes by default.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Added the read-one case in `3c78758` — `404` unless `includeDeleted=true`, which keeps soft deletion
invisible by default and makes the parameter behave the same on both read shapes.

Leaving `PUT` and actions against a soft-deleted resource to the owning team. Whether that's `404`,
`409`, or an implicit restore depends on what the resource means, and picking one here would be
guessing on their behalf.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+| -------------------- | --------------------------------------------------- |
+| `var`                | Reference a field of the resource.                  |
+| `==`, `!=`           | Equal, not equal.                                   |
+| `>`, `>=`, `<`, `<=` | Numeric and date comparison.                        |
+| `in`                 | Membership in a list, or a substring match.         |
+| `and`, `or`          | Boolean composition, each taking two or more terms. |
+| `!`                  | Negation.                                           |
+
+A request that uses any other operator, or references a field that does not exist, `MUST` be
+rejected with `422`.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Good catch — aligned in `4cd82cc`. Line 771 now reads "references a field the API does not support,"
matching the `filter[{name}]` rule. The search body exists to express combinations `filter[…]` can't,
not to widen what's queryable, so a caller shouldn't reach unfilterable attributes by changing
transport. It also strictly subsumes the old test, since a field that doesn't exist isn't supported
either.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+1. Changing the datatype of a field.
+1. Making an optional field required.
+1. Adding additional constraints to a field.
+
+The lone exception to this rule are security and cryptographic issues. Such issues `SHOULD` be fixed
+"in place", without creating new versions of the API, even if the changes would technically be
+considered breaking changes.
+
+Otherwise if changes need to be made that _would_ be breaking changes, a new version of the API must
+be created and the old one [deprecated](#deprecation).
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Most of this is superseded — the section was rewritten in `7a4da3b` and `4acc040` in response to
reviewer threads. "The lone exception to this rule" is gone, the bug carve-out is restored and now
covers any bug rather than only added constraints, and line 127 states why the verb is `SHOULD NOT`
rather than leaving it unexplained.

Point 3 was real and is fixed in `10c2070`: the bare lowercase "must" is now `MUST`. It was the only
unmarked normative word in the file.

Not reverting to `MUST NOT`. The softening was requested by a reviewer and is deliberate — these are
internal APIs where we own every caller, and the `MAY` and bug paragraphs channel the judgment rather
than leaving it open-ended.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+duplicated in the URL path.
+
+## Breaking changes and versioning
+
+APIs `SHOULD NOT` make **breaking changes**. A breaking change is defined as follows:
+
+1. Adding a new, required field.
+1. Making an optional field required.
+1. Changing the datatype of a field.
+1. Adding additional constraints to a field.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Agreed, and fixed in `0d4e7db`. The definition was request-side only, which left response-field
removal uncovered — and you're right that Deprecation is operation-level, so there was no announced
path for it.

The old item was cut because it didn't distinguish direction: removing a _request_ field isn't
breaking, since unrecognized fields are ignored. Removing a _response_ field is. So the definition now
covers both directions explicitly and the list has "Removing a field from a response" rather than the
old undirected item.

Not adopting the ±3 scoping — that mechanism assumes a fleet-wide version coordinate, which
independently versioned services don't have.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1051 @@
...
+caller. A team `MAY` make a breaking change in place when it can account for every caller — either
+because the change has been coordinated with them, or because the rejection surfaces somewhere the
+caller or the user can act on it.
+
+The lone exception to this rule are security and cryptographic issues. Such issues `SHOULD` be fixed
+"in place", without creating new versions of the API, even if the changes would technically be
+considered breaking changes.
+
+Otherwise if changes need to be made that _would_ be breaking changes, a new version of the API must
+be created and the old one [deprecated](#deprecation).
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Superseded — all three points are closed by commits after this review.

"The lone exception to this rule" is gone as of `4acc040`, replaced by "Bugs, however, `SHOULD` be
fixed in place," which also resolves the number agreement. The lowercase "must" became `MUST` in
`10c2070`. With "lone" removed, the `Otherwise` at line 137 scopes correctly: `SHOULD NOT` in general,
`MAY` in place when every caller is accounted for, `SHOULD` in place for bugs, `MUST` version once
none of those apply. Your trace now lands in the `MAY` paragraph with nothing contradicting it.

Not taking the suggested rewrite: "A team that cannot account for every caller `MUST` create a new
version" predates the bug paragraph, and would force a bug into a new version whenever callers can't
be accounted for — contradicting line 134. `Otherwise` works because it chains off both preceding
paragraphs instead of naming one condition.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1050 @@
...
+1. Changing the datatype of a field.
+1. Adding additional constraints to a field.
+
+The verb is `SHOULD NOT` rather than `MUST NOT` because these are internal APIs and we own every
+caller. A team `MAY` make a breaking change in place when it can account for every caller — either
+because the change has been coordinated with them, or because the rejection surfaces somewhere the
+caller or the user can act on it.
+
+Bugs, however, `SHOULD` be fixed "in place", without creating new versions of the API, even if the
+changes would technically be considered breaking changes.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Declining. This is a deliberate change, made in response to a reviewer thread asking for exactly it.

On "too narrow": a vulnerability is a defect, so it's a bug and it's covered. The hardening case you
cite may or may not be a defect, and where it isn't, the paragraph above covers it — a team `MAY` make
a breaking change in place when it can account for every caller. Neither path leaves a vulnerable
version in service.

On "too broad": that's a fair reading and it's the cost we took knowingly. "Bug" is broad, and
"intended but not enforced" could be stretched. These are internal APIs where we own every caller, and
we would rather trust teams and keep this section short than enumerate qualifying classes of defect.
If we see it stretched, narrowing it is available.

Line 136 is already `MUST` as of `10c2070`.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1056 @@
...
+APIs that support specifying multiple values for a query parameter `MUST` do so via a
+comma-delimited list. This allows APIs that started with single-value query parameters to evolve to
+supporting multiple-values without changing the external-facing contract. When multiple values are
+provided, "or" query semantics `MUST` be applied.
+
+A comma that is part of a value `MUST` be escaped with a backslash —
+`filter[displayName]=Smith\, Jr` is the single value `Smith, Jr`. This applies only where the caller
+is supplying multiple values or a [range](#ranges); a parameter that takes a single value is never
+split, so a comma in its value needs no escaping. An asterisk that is part of a value `MUST`
+likewise be escaped wherever the API supports wildcard matching.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Two of the three are fixed in `cf0be8a`; the scoping stays.

- The circular phrasing is gone — it now reads "On a parameter that accepts multiple values or a
  range," which is a property of the API rather than of caller intent.
- The escape character is now escapable: a backslash that is part of a value `MUST` be escaped, so
  `C:\,D:\` is no longer ambiguous.

On making escaping unconditional: declining. That would oblige every caller of every parameter to
escape commas forever, to guard against an evolution that happens occasionally and almost always on
constrained fields — `filter[color]=RED` becoming `filter[color]=RED,BLUE` — whose values cannot
contain commas in the first place. The cost lands on every API that will ever exist; the benefit is
concentrated in a narrow case.

Instead the hazard is now advice to the owner making that change:

> **Take care when a single-value parameter gains multi-value support.** A caller sending an
> unescaped comma will suddenly be sending two values. That is usually safe, because the evolution
> typically happens on constrained fields whose values cannot contain commas — but confirm it before
> making the change.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1064 @@
...
+
+APIs `SHOULD NOT` make **breaking changes**. A breaking change is any change that causes a request
+that is valid today to be rejected tomorrow, or a response that a caller can process today to become
+unprocessable. In practice, that means:
+
+1. Adding a new, required field.
+1. Making an optional field required.
+1. Changing the datatype of a field.
+1. Adding additional constraints to a field.
+1. Removing a field from a response.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: The scenario is real and the trace is right — marking a field `nullable` interacts with line 176 so
that the field is omitted outright rather than sent as `null`, and being partial does make it harder
to catch than a removal.

Declining the change anyway, because the definition already covers it: "a response that a caller can
process today to become unprocessable" is precisely what this is, so a team applying the rule as
written reaches the right answer without the list naming the case. The numbered items are the common
cases at a glance, not an exhaustive taxonomy — enumerating every route to the same failure is how a
standard stops being read. Comprehensibility is worth more to us here than completeness.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1064 @@
...
+| `200 OK`       | The resources were updated.                | `meta.affectedCount`, the number of resources updated.       |
+| `202 Accepted` | The resources are scheduled to be updated. | The [job](#jobs) that was scheduled to update the resources. |
+
+## Bulk deletes
+
+Bulk deletes delete every resource matching a filter.
+
+- HTTP verb `MUST` be `POST`.
+- API path `MUST` be like `/api/v1/{resource plural}:bulk-delete`.
+- `filter` `MUST` be present and use the [advanced query](#advanced-queries) expression language.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fair — "Delete APIs" is ambiguous as to whether a `POST`-shaped bulk operation counts, so I've added
one line in `58ead17`: on a service that supports soft deletes, a bulk delete `MUST` soft-delete.

Leaving out the `permanent=true` half. That's a query parameter and bulk delete carries a body, so
supporting it means inventing transport for it — and a filter-driven irreversible mass delete isn't a
capability we want to pave a road to. A service that needs it can define it.

On bulk update and `version`: correct, but inherent to filter-driven updates rather than an oversight.
There's visibly nowhere to put a version, so we're not worried about a team trying to reconcile the
two.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1064 @@
...
+- Request and responses use the `application/json` media type, not `application/vnd.api+json`.
+- We do not support [compound documents](https://jsonapi.org/format/#document-compound-documents) or
+  the related `include` query parameter.
+- Paged responses are not required to include `links`.
+- For sparse fieldsets, callers specify the fields to include via a `fields` query parameter. We do
+  not support the `fields[{type}]` form.
+- Unrecognized query parameters are ignored rather than rejected with `400`, and the result-shaping
+  subset is rejected with `422`. See
+  [Unrecognized fields, query parameters, and headers](#unrecognized-fields-query-parameters-and-headers).
+- Implementation-specific query parameters are not required to contain a non-`a-z` character.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Both added in `3a4f123`. These are the same category as the last two — choices already made, just not
recorded where the ADR promised the exceptions would be listed.

You're right that `relationships` is the more consequential of the two. It's named in the ADR's cons
but nowhere on this page, and line 41 tells readers to assume the specification where we're silent —
so someone reaching for a client library would have found no signal at all.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1068 @@
...
+
+APIs that support specifying multiple values for a query parameter `MUST` do so via a
+comma-delimited list. This allows APIs that started with single-value query parameters to evolve to
+supporting multiple-values without changing the external-facing contract. When multiple values are
+provided, "or" query semantics `MUST` be applied.
+
+On a parameter that accepts multiple values or a [range](#ranges), a comma or backslash that is part
+of a value `MUST` be escaped with a backslash — `filter[displayName]=Smith\, Jr` is the single value
+`Smith, Jr`. A parameter that takes a single value is never split, so its value needs no escaping.
+An asterisk `MUST` likewise be escaped wherever the API supports wildcard matching.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Right on both counts, and both were mine — fixed in `eff826e`.

Dropping "a comma in" from the exemption is what broadened it across the asterisk rule, and you're
right that wildcard support is a property of the API rather than of arity, so the two sentences
collided. The backslash clause is now tied to wherever either escape applies rather than to arity,
which closes `C:\*` the same way `cf0be8a` closed `C:\,D:\`.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1074 @@
...
+| `202 Accepted`   | The resource is scheduled to be updated. | The [job](#jobs) that was scheduled to update the resource. |
+| `204 No Content` | The resource was updated.                | Nothing.                                                    |
+
+See also: [Partial updates](#partial-updates)
+
+## Partial updates
+
+Services `SHOULD NOT` support partial updates — see [the FAQ](#frequently-asked-questions) for why.
+A service that _does_ support them `MUST` use `PATCH` and `MUST` ignore any field that isn't
+specified.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Real contradiction — fixed in `e3e3a75`. The `PATCH` section now reads:

> A service that _does_ support them `MUST` use `PATCH` and `MUST` ignore any field that isn't
> specified. In a partial update, an absent field means "do not change" and an explicit `null` clears
> the value.

Scoped there rather than on the JSON rule, since `PATCH` is the exception and we prefer to state
exceptions where the topic is discussed. That also closes your second point: an explicit `null` is now
the documented way to clear a field in a partial update, so `""` collapsing to "not present" no longer
leaves it unclearable.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1074 @@
...
+
+**Success responses**
+
+Upon success, APIs `SHOULD` return `200` but `MAY` return `202` or `204`.
+
+| Status           | Description                              | Response                                                    |
+| ---------------- | ---------------------------------------- | ----------------------------------------------------------- |
+| `200 OK`         | The resource was updated.                | The latest representation of the resource.                  |
+| `202 Accepted`   | The resource is scheduled to be updated. | The [job](#jobs) that was scheduled to update the resource. |
+| `204 No Content` | The resource was updated.                | Nothing.                                                    |
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: The missing status was real — `201 Created` added to the update success table in `e3e3a75`, and the
prose now reads "`SHOULD` return `200` — or `201` if the request created the resource." A caller can
now tell whether it created or replaced. Added only to the `PUT` table, not the `PATCH` one, since a
partial update can't create.

Declining the other two points:

- `POST` vs `PUT` isn't a conflict. "Creating resources" describes `POST /api/v1/{resource plural}`;
  create-or-update is `PUT /api/v1/{resource plural}/{id}`. Different operations, and line 279 already
  sanctions the upsert.
- On `version` for a `PUT` that creates: there's no stored version to echo, so there's nothing for a
  caller to reconcile.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1074 @@
...
+| `filter[failedLoginCount]=[3,*)`                                | Users who have 3 or more failed login attempts. |
+| `filter[birthDate]=(*,2000-01-01T00:00:00Z)`                    | Users born before the year 2000.                |
+| `filter[birthDate]=[2000-01-01T00:00:00Z,2001-01-01T00:00:00Z)` | Users born in the year 2000.                    |
+
+See also [Advanced Queries](#advanced-queries).
+
+## Paging
+
+APIs `SHOULD` support paging and those that do `MUST` support either **offset/limit** style paging
+or **cursor** style paging.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Accurate, but declining. This is the same question we settled earlier in this PR when we struck the
requirement that services declare a maximum number of resources a single request may affect: these
are internal-only APIs and we trust teams to work limits like that out for their own resources. A
mandatory declared maximum for `page[size]` is that requirement in a different section, and the
expression-size bound you suggest for advanced queries is the same idea a third time.

Teams remain free to impose and document their own bounds — the request-validation rules already
cover enforcing them once declared.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1077 @@
...
+Upon success, APIs `MUST` return `200`.
+
+| Status   | Description             | Response                |
+| -------- | ----------------------- | ----------------------- |
+| `200 OK` | Request was successful. | The resource requested. |
+
+## Updating resources
+
+- HTTP verb `MUST` be `PUT`.
+- API path `SHOULD` be like `/api/v1/{resource plural}/{id}`.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Closing without changes. The gap is real — nothing here forces the body `id` to match the path, and
the authorization variant you describe is a legitimate concern.

We're declining on scope. This standard cannot enumerate every ambiguity it leaves open without
becoming a document nobody reads end to end, and request binding and authorization are things the
owning service already has to get right. Matching the path is the obvious implementation; we'd rather
trust teams with it than spend the rule.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1077 @@
...
+
+**Success responses**
+
+Upon success, APIs `SHOULD` return `201` but `MAY` return `202` or `204`.
+
+| Status           | Description                          | Response                                                    |
+| ---------------- | ------------------------------------ | ----------------------------------------------------------- |
+| `201 Created`    | Resource was created.                | The latest representation of the resource.                  |
+| `202 Accepted`   | Resource is scheduled to be created. | The [job](#jobs) that was scheduled to create the resource. |
+| `204 No Content` | Resource was created.                | Nothing.                                                    |
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Closing without changes. A create that returns `204` for a server-generated ID leaves the caller with
nothing, which is self-evidently unhelpful — we'd rather trust teams not to do it than spend a table
cell forbidding it.

Declining the `Location` requirement as well: `201` already returns the representation including the
`id`, so the header would be a new obligation on every create for little gain.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1077 @@
...
+| `filter[failedLoginCount]=[3,*)`                                | Users who have 3 or more failed login attempts. |
+| `filter[birthDate]=(*,2000-01-01T00:00:00Z)`                    | Users born before the year 2000.                |
+| `filter[birthDate]=[2000-01-01T00:00:00Z,2001-01-01T00:00:00Z)` | Users born in the year 2000.                    |
+
+See also [Advanced Queries](#advanced-queries).
+
+## Paging
+
+APIs `SHOULD` support paging and those that do `MUST` support either **offset/limit** style paging
+or **cursor** style paging.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Closing without changes, though this is the strongest of the four and the distinction you draw from
the page-size question is a fair one.

The reason is scope rather than disagreement. Stable ordering under paging is something any competent
implementation has to solve, and a team building a paged read against a real store will meet it
immediately. We're choosing to keep this document short enough to be read rather than complete enough
to be cited for every failure mode.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1077 @@
...
+
+## Advanced queries
+
+The `filter` query parameters cover simple predicates, but they cannot express arbitrary `AND` and
+`OR` combinations, and a long expression will exceed the practical limit on URL length. APIs that
+need richer queries `MUST` expose a **search** operation that carries the expression in the request
+body.
+
+- HTTP verb `MUST` be `POST`.
+- API path `MUST` be like `/api/v1/{resource plural}:search`.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Closing without changes. The existing usage already makes the split clear — every colon operation in
this document is collection-level and every `actions` path carries an `{id}` — so a team has a working
pattern to follow without us writing the rule down.

docs/architecture/adr/0036-internal-api-standards.md
quexten: > What we want to avoid is a situation where an API is changed, requests start getting rejected, and the user has no way to fix it (i.e. to get the client or UI they are using to "do the right thing"). But
if I click save one day and get an error telling me something isn't valid and needs correcting, and
the client/UI lets me fix it, that is perfectly fine.


In this case the requests are not valid compared to our intention, but valid as in "accepted by the server today". But these already do not lead to valid data. [specifically we are adding guards to prevent certain kinds of client bugs using wrong keys resulting in undecryptable vault items, via server validation of metadata].

So i guess valid here means "as compared to our intention" / needs to not obviously break the clients for users, which our use case does not. Thanks, that is very helpful!!

> So for the blob cases specifically: tightening what a vault item data blob or an encryption key field
will accept is a breaking change a team may make in place, provided the rejection surfaces somewhere the user or the calling service can act on it. If it would instead surface as a silent sync failure with no remediation path, that is the case the standard is trying to prevent — and then it does belong behind a new version.

I want to drill in on this a bit more. Suppose we add a new type of encryption today. Existing clients will not understand what this data field means. The data shape did not change, but the set of legal values when *reading* data expanded. I know we consider *narrowing* when *writing* a breaking change, but do we consider *widening* when *reading* a breaking change? Since existing clients would not understand this data type?

How would we deal with this? Would we have to add client support code to ignore unknown values encrypted with this vault item type?

What if the encryption keys of the users themselves change to a type the client does not support yet, or other cryptographic operations are necessary that the client doesn't support yet / if the actions they can do are available on the API level, but not permitted by the cryptographic format version of their account?

If we no longer define a set of supported client versions, I'm not sure I understand how to deal with this situation.

docs/architecture/server/internal-api-standards.md
quexten: The way we have dealt with this in the past is "Roll out support for a new feature / format first, then a few "versions" later start migrating to the feature so that most clients support the feature".

docs/architecture/adr/0036-internal-api-standards.md
mike-gorman-bitwarden: > If we no longer define a set of supported client versions, I'm not sure I understand how to deal
> with this situation.

We do still define them.  Nothing in these standards changes that.  These standards are for internal,
service-to-service APIs only.  Mobile, CLI, and browser clients don't call internal APIs, so whatever
client support policy we have today remains untouched.

Regarding "widening when reading", I don't think that one is an API versioning question at all.  The
API says the field is a `string`.  After you add a new encryption type it is still a `string`, it
still deserializes, and it still passes validation.  What changed is the _meaning_ of the blob.  As
far as the API is concerned, it's just bytes.

So versioning the API wouldn't make a difference because the client still can't decrypt the value once
it has it.

I think it is a data format versioning question.  From what I have seen, `EncString` carries its own
type and `EncryptionType` is already up to 7 (`CoseEncrypt0`), so we have done exactly what you are
describing at least once without needing a new API version.  A client that runs into a type it doesn't
recognize knows it doesn't recognize it, because the type is right there in the data. What the client
does about that is a crypto question, not an API question.

Regarding account-level crypto versions, I don't think API versioning helps there, either.  If the
account is in a format the client can't handle, the data is in that format no matter which endpoint
the client calls.  The client has to look at the account's crypto version and tell the user to update
rather than try and make a mess of it.

Definitely an issue but it is more of a client capability question, I think, and less of an API
standards issue.

docs/architecture/server/internal-api-standards.md
```diff
@@ -0,0 +1,1078 @@
...
+
+# Internal API standards
+
+{/* cspell:ignore reate reates eletes elete pdate pdates fieldsets */}
+
+**Audience:** Bitwarden engineers and AI agents building or consuming a service-to-service API.
+
+**Scope.** Internal, service-to-service APIs. Bitwarden's existing public API, including the part
+that is currently referred to as "internal", is out of scope and is not changing. This ADR applies
+only to internal, service-to-service APIs that are not exposed to the internet.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Fixed.

docs/architecture/server/service-to-service-api-standards.md
```diff
@@ -0,0 +1,1078 @@
...
+> **Take care when a single-value parameter gains multi-value support.** A caller sending an
+> unescaped comma will suddenly be sending two values. That is usually safe, because the evolution
+> typically happens on constrained fields whose values cannot contain commas — but confirm it before
+> making the change.
+
+### Ranges
+
+Query parameters that allow the caller to specify a range of values should do so using two values,
+separated by commas, and using `[` and `]` to represent **inclusive** begin and end, and `(` and `)`
+to represent **exclusive** begin and end. An asterisk `*` may be used to represent no limit.
```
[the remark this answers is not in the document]
mike-gorman-bitwarden: Both fixed in fa49b45.

Ranges now reads `MUST` do so using two values … an asterisk `*` `MAY` be used to represent no limit — agreed that a wire grammar two services have to agree on byte-for-byte cannot be the one section a reader is entitled to treat as advisory, especially with the comma-escape rule already normative against it.

Line 170 taken as well: the all-caps enum convention is now `SHOULD`, which also matches the whitespace-stripping bullet two lines down.

docs/architecture/adr/0036-service-to-service-api-standards.md
```diff
@@ -0,0 +1,182 @@
...
+status: Proposed
+date: 2026-09-22
+tags: [server, server-sdk]
+---
+
+# 0036 - Service-to-service API standards
+
+<AdrTable frontMatter={frontMatter}></AdrTable>
+
+{/* cspell:ignore Appibase Catalio */}
```
withinfocus: ❌ We just maintain a list of accepted words in the shared config file in the root.

docs/architecture/adr/0036-service-to-service-api-standards.md
```diff
@@ -0,0 +1,182 @@
...
+This ADR uses [RFC 2119](https://www.rfc-editor.org/info/rfc2119/) keywords (`MUST`, `MUST NOT`,
+`SHOULD`, `SHOULD NOT`, `MAY`) deliberately. Anything marked `MUST` or `MUST NOT` is not negotiable
+at team level; a team that needs an exception brings the case to the architecture group.
+
+**Service-to-service API** means one Bitwarden service calling another. Bitwarden's existing public
+API is out of scope and is not changing.
+
+## Context and problem statement
+
+Service-to-service APIs are new. As server functionality is decomposed into independently deployable
```
withinfocus: 🎨 I recommend softening this, because we have this in place today already, with the mothership.

docs/architecture/server/service-to-service-api-standards.md
```diff
@@ -0,0 +1,1079 @@
+---
```
withinfocus: ❌ This can come with the ADR acceptance.

withinfocus (changes requested):

docs/architecture/adr/0036-internal-api-standards.md
mike-gorman-bitwarden: ADR is now generalized.  The specific standards will be iterated on externally.

docs/architecture/adr/0036-service-to-service-api-standards.md
```diff
@@ -0,0 +1,181 @@
...
+The specific standards, including where we deviate from JSON:API, are published here:
+[Internal API standards](../server/internal-api-standards.md). That page is the living reference:
+its rules evolve by pull request without superseding this decision, and this ADR is superseded only
+if the model itself changes.
+
+### Positive consequences
+
+- One document shape across every internal API, so a consumer calling five services learns one error
+  format, one pagination scheme, and one filter grammar.
+- Versioning gives a contract a way to change shape. A breaking change becomes a new version with a
```
mike-gorman-bitwarden: ADR is now generalized.  The specific standards will be iterated on externally.

docs/architecture/server/service-to-service-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+duplicated in the URL path.
+
+## Breaking changes and versioning
+
+APIs `MUST NOT` make **breaking changes**. A breaking change is defined as follows:
+
+1. Removing (or relocating) a field.
+1. Changing the datatype of a field.
+1. Making an optional field required.
+1. Adding additional constraints to a field.
```
mike-gorman-bitwarden: ADR is now generalized.  The specific standards will be iterated on externally.

docs/architecture/adr/0036-service-to-service-api-standards.md
mike-gorman-bitwarden: Done. bad08b9690772fefdf012f34ea5c8614be61931c

docs/architecture/server/service-to-service-api-standards.md
mike-gorman-bitwarden: Removed.

docs/architecture/adr/0036-internal-api-standards.md
mike-gorman-bitwarden: ADR is now generalized. The specific standards will be iterated on externally.

docs/architecture/server/service-to-service-api-standards.md
```diff
@@ -0,0 +1,1045 @@
...
+> adding one.
+
+## Unrecognized fields, query parameters, and headers
+
+Sometimes a new, optional field, query parameter, or header is added to an API. Consumers upgrade to
+new clients and start passing the new fields/parameters/headers and everything is great until a
+security issue is discovered. The service is rolled back to the previous version but consumers are
+still using newer versions of the client.
+
+To keep from having to _also_ revert all of the consumers, APIs `MUST` ignore unrecognized fields,
```
mike-gorman-bitwarden: ADR is now generalized. The specific standards will be iterated on externally.

docs/architecture/adr/0036-service-to-service-api-standards.md
mike-gorman-bitwarden: Removed.

withinfocus (approved):

PR #866, opened 2026-09-16
Document why UniFFI exports must return BitwardenError

dani-garcia:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C054ZQSBS49/p1789450329988709
https://github.com/bitwarden/sdk-internal/pull/1486

## 📔 Objective

The feature clients require a wrapper for UniFFI, while they don't for WASM. This is due to a UniFFI limitation on error handling when failing to parse arguments. Update the wrappers section to add a bit more detail on why we do this.

While at it, I've noticed that some of the code samples in this section didn't quite match what we actually do in the SDK (Returning the clients wrapped in `Arc`, `SharedClient`). I've updated those samples to more closely match reality. Also updated a couple file links that have since been moved.


docs/architecture/sdk/adding-functionality.md
```diff
@@ -165,38 +165,41 @@ If the new functionality needs to be available on mobile platforms (Android / iO
...
     }
 }
 ```
 
 ### Create the wrapper
 
-Create a wrapper struct that holds the SDK `Client` and delegates to the underlying Rust client. See
-[`bitwarden-uniffi/src/tool/sends.rs`][uniffi-sends] for a complete example.
+Create a wrapper struct that holds the feature crate's client and delegates to it. See
+[`bitwarden-uniffi/src/tools/sends.rs`][uniffi-sends] for a complete example.
```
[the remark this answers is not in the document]
dani-garcia: good point, updated it

coltonhurst (approved):

docs/architecture/sdk/client-patterns.md
```diff
@@ -44,31 +44,45 @@ If the client will be exposed over WASM, annotate both the struct and its `impl`
...
-wrappers through accessor methods, just like the application interface clients do. For example, a
-`VaultClient` wrapper returns `Arc<FoldersClient>`.
+The wrapper mirrors the structure of the Rust client hierarchy: parent wrappers expose child
+wrappers through accessor methods, just like the application interface clients do. `VaultClient`
+wraps `bitwarden_vault::VaultClient`, and its `folders()` accessor returns the `FoldersClient`
+wrapper shown above.
+
+#### Why the wrappers exist
+
+UniFFI reports a failed argument conversion by downcasting to the error type declared in the
```
quexten, line 69: A question: Could the BitwardenError type be moved out to a separate crate, so that the clients don't have to live in bitwarden-uniffi?

quexten (approved):

docs/architecture/sdk/client-patterns.md
Hinton: You can replicate the same pattern outside of bitwarden-uiffi, but not sure when you would want to use other clients since we generally want to have a single SDK client with sub-clients.

Hinton (approved):

